PR #5, opened 2016-08-22
Change signature of dispatchEvent to take in LogEvent object rather than url and request parameters

vraja2:
Since the v2 endpoint uses POST requests, I'm reintroducing the `LogEvent` object to generalize request logic. Also creating a base `EventBuilder` from which `EventBuilderV1` and in the future, `EventBuilderV2` will inherit.

@aliabbasrizvi @jophde @haleybash-optimizely @delikat @alda-optimizely @wangjoshuah


core-api/src/main/java/com/optimizely/ab/event/NoopEventHandler.java
```diff
@@ -29,7 +29,8 @@
     private static final Logger logger = LoggerFactory.getLogger(NoopEventHandler.class);
 
     @Override
-    public void dispatchEvent(String url, Map<String, String> params) {
-        logger.debug("Called dispatchEvent with URL: {} and params: {}", url, params);
+    public void dispatchEvent(LogEvent logEvent) {
```
aliabbasrizvi: May be not change this signature and have `dispatchEvent` continue having the same signature, with the consumer providing this information. `dispatchEvent` should be as naive as possible with minimum overhead for anyone trying to define it. In this case, they would need to know the methods provided by `LogEvent` to be able to create `dispatchEvent`

vraja2: Passing a `Map<String, String>` makes it so that you must perform explicit serialization rather than using an existing library like Jackson or Gson (can't perform type inference without POJOs), so we'll have typecasts from `String` for each POST parameter. This isn't an issue with untyped languages since a dictionary can store multiple types, but in Java it's difficult to generate a payload string from a `Map` and the convention is to serialize from a POJO.  I think using `LogEvent` has a low learning curve since it doesn't contain any Optimizely-specific logic.

jophde: From my POV as a consumer of this API I don't see much anymore complexity in dealing with LogEvent or a raw String and map.  Auto complete makes it pretty easy to see that LogEvent has a method for getting the URL and getting the params. Down the road it might be useful to add more methods to it like one that returns a map of headers and header values. 

It also seems useful to put the RequestMethod on LogEvent.  I am not sure I would have known to use GET for the old event API if I didn't work at Optimizely.

Do we really need to the EventApiV1.. abstraction?  Wouldn't the only thing that ever changes be the endpoint and params?  These are internal to the EventApi class so changing those wouldn't break public API compatibility.

core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilderV1.java
```diff
@@ -0,0 +1,274 @@
...
+                                          @Nonnull Variation variation,
+                                          @Nonnull String userId,
+                                          @Nonnull Map<String, String> attributes) {
+
+        Map<String, String> requestParams = new HashMap<String, String>();
+        addCommonRequestParams(requestParams, projectConfig, userId, attributes);
+        addImpressionGoal(requestParams, activatedExperiment);
+        addExperiment(requestParams, activatedExperiment, variation);
+
+        return new LogEvent(RequestMethod.GET,
```
wangjoshuah: Instead of passing around the requestParams dictionary we should just use createImpressionEvent in the LogEvent class.

wangjoshuah:
yeah specific properties of a POJO are super easy to work with since your IDEA gets them for you automatically.

core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilderV1.java
vraja2: We should make LogEvent as lightweight as possible since developers will be using it and we want to abstract event creation away from them. In a future change, instead of passing around a `Map` we can use `putAll` to combine maps returned by helper functions in the root creation method.

aliabbasrizvi:
LGTM

PR #8, opened 2016-08-25
Add EventBuilderV2

vraja2:
- Add `layerId` field to `Experiment`
- Add `EventBuilderV2` and conditionally instantiates based on datafile version
- Fix some tests

@aliabbasrizvi @delikat @haleybash-optimizely @mikeng13 @alda-optimizely @wangjoshuah @jophde


core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -160,8 +161,8 @@ private Optimizely(@Nonnull ProjectConfig projectConfig,
         LogEvent impressionEvent =
                 eventBuilder.createImpressionEvent(projectConfig, experiment, variation, userId, attributes);
         logger.info("Activating user \"{}\" in experiment \"{}\".", userId, experiment.getKey());
-        logger.debug("Dispatching impression event to URL {} with params {}.", impressionEvent.getEndpointUrl(),
-                     impressionEvent.getRequestParams());
+        logger.debug("Dispatching impression event to URL {} with params {} and payload \"{}\".",
+                     impressionEvent.getEndpointUrl(), impressionEvent.getRequestParams(), impressionEvent.getBody());
```
aliabbasrizvi: How is adding payload helping? Wouldn't there be a lot of overlap between params and body already?

core-api/src/main/java/com/optimizely/ab/config/Experiment.java
```diff
@@ -56,19 +57,21 @@
     public Experiment(@JsonProperty("id") String id,
                       @JsonProperty("key") String key,
                       @JsonProperty("status") String status,
+                      @JsonProperty("layerId") String layerId,
                       @JsonProperty("audienceIds") List<String> audienceIds,
                       @JsonProperty("variations") List<Variation> variations,
                       @JsonProperty("forcedVariations") Map<String, String> userIdToVariationKeyMap,
                       @JsonProperty("trafficAllocation") List<TrafficAllocation> trafficAllocation) {
-        this(id, key, status, audienceIds, variations, userIdToVariationKeyMap, trafficAllocation, "");
+        this(id, key, status, layerId, audienceIds, variations, userIdToVariationKeyMap, trafficAllocation, "");
```
aliabbasrizvi: This ID i.e. `layerId` will not be there in the version 1 JSON file. Just FYI

core-api/src/main/java/com/optimizely/ab/Optimizely.java
vraja2: At any point, exclusively payload or params will be populated (get vs post requests). I'm including both because it's not worth slowing down the SDK by having a conditional to check whether we're sending a V1 or V2 event. Thus, the empty variable will be printed as "{}"

core-api/src/test/resources/config/no-audience-project-config.json
```diff
@@ -1,13 +1,14 @@
 {
   "accountId": "789",
   "projectId": "1234",
-  "version": "4",
+  "version": "2",
```
aliabbasrizvi: We still need to continue supporting version 1. We should have separate files and make sure everything works.

aliabbasrizvi:
I think you will have to rework some of the logic here to be able to continue supporting v1 and v2 data file changes which is limited to layer ID.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
jophde: How expensive would this check really be?  It's just a constant time operation to check whether or not `eventBuilder instanceof EventBuidlerV2` right?  Printing out an empty `{}` seems like it would be confusing to someone reading the logs.

jophde:
Looks pretty good to me.  How many people do we actually have running experiments using v1 events?  I question whether it's worth even back supporting this since we are still in Beta.  Why convert data files to v2 on the backend if we must?  Seems like a lot of logic to add to the SDK.  After we release > 99% of customers will be using v2 endpoints going forward.

Since nobody is on the Android event handler yet can I just make it use v2 without back support?

core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilderV2.java
```diff
@@ -0,0 +1,187 @@
...
+
+    public LogEvent createImpressionEvent(@Nonnull ProjectConfig projectConfig,
+                                          @Nonnull Experiment activatedExperiment,
+                                          @Nonnull Variation variation,
+                                          @Nonnull String userId,
+                                          @Nonnull Map<String, String> attributes) {
+
+        Impression impressionPayload = new Impression();
+        impressionPayload.setVisitorId(userId);
+        impressionPayload.setTimestamp(System.currentTimeMillis());
```
vraja2: yes

vraja2:
@aliabbasrizvi Addressed your feedback

core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilderV2.java
```diff
@@ -0,0 +1,192 @@
...
+        impressionPayload.setIsGlobalHoldback(false);
+        impressionPayload.setProjectId(projectConfig.getProjectId());
+
+        Decision decisionTicket = new Decision();
```
alda-optimizely: Just to avoid confusion...this object is a Decision, not a Decision Ticket.

vraja2: yeah that's correct, I'll fix this. good catch

core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilderV2.java
```diff
@@ -0,0 +1,192 @@
...
+        Decision decision = new Decision();
+        decision.setVariationId(variation.getId());
+        decision.setIsLayerHoldback(false);
+        decision.setExperimentId(activatedExperiment.getId());
+        impressionPayload.setDecision(decision);
+
+        impressionPayload.setLayerId(activatedExperiment.getLayerId());
+        impressionPayload.setAccountId(projectConfig.getAccountId());
+        impressionPayload.setUserFeatures(createFeatures(attributes, projectConfig));
+
```
haleybash-optimizely: Could you also add `clientEngine` and `clientVersion`?

core-api/src/main/java/com/optimizely/ab/config/Experiment.java
haleybash-optimizely: Would you want to add a comment that emphasizes `layerId` is for v2 JSON datafile only?

core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilderV2.java
```diff
@@ -0,0 +1,192 @@
...
+    private List<Feature> createFeatures(Map<String, String> attributes, ProjectConfig projectConfig) {
+        Map<String, Attribute> attributeKeyMapping = projectConfig.getAttributeKeyMapping();
+        List<Feature> features = new ArrayList<Feature>();
+
+        for (Map.Entry<String, String> attributeEntry : attributes.entrySet()) {
+            String attributeKey = attributeEntry.getKey();
+            Attribute attribute = attributeKeyMapping.get(attributeKey);
+
+            if (attribute == null) {
+                logger.warn("attempting to use unknown attribute key: {}. Attribute will be ignored", attributeKey);
```
haleybash-optimizely: micronit: could you capitalize Attempting?

core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilderV2.java
```diff
@@ -0,0 +1,192 @@
...
+        }
+
+        return features;
+    }
+
+    /**
+     * Helper method to create {@link LayerState} objects for all experiments mapped to an event.
+     * <p>
+     * Note, although the user may not have <i>activated</i> all experiments, we include all <i>possible</i> experiment
+     * mappings so that  activated experiment state doesn't have to be maintained and passed around server-side.
```
aliabbasrizvi: nit. 2 spaces between that and activated

core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilderV2.java
```diff
@@ -0,0 +1,192 @@
...
+     * <p>
+     * For example, in a project with multiple experiments, if user 'user1' activates experiment 'exp1' then converts,
```
aliabbasrizvi: nit. and then converts

core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilderV2.java
```diff
@@ -0,0 +1,192 @@
...
+     * we need to ensure we attribute that conversion to 'exp1'. We can either do that by recording activation
+     * and passing that state around so it's available on conversion, or we can send all possible activations at
+     * conversion-time and attribute conversions <i>only</i> to the experiments for which we also received an
+     * impression.
+     * <p>
+     * This is referred to as "visitor first counting". It's important to filter the bucket map as much as possible,
```
aliabbasrizvi: No need for this para

core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilderV2.java
```diff
@@ -0,0 +1,192 @@
...
+     * as any experiments sent through that don't have a corresponding impression are simply occupying space for
+     * no good reason.
+     *
+     * @param projectConfig the current project config
+     * @param bucketer the bucketing algorithm to use
+     * @param userId the user's id for the impression event
+     * @param goalKey the goal that the bucket map will be filtered by
```
aliabbasrizvi: Lets just call it eventKey now

core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilderV2.java
```diff
@@ -0,0 +1,192 @@
...
+            String attributeKey = attributeEntry.getKey();
+            Attribute attribute = attributeKeyMapping.get(attributeKey);
+
+            if (attribute == null) {
+                logger.warn("Attempting to use unknown attribute key: {}. Attribute will be ignored", attributeKey);
+                continue;
+            }
+
+            features.add(new Feature(attribute.getId(), attributeKey, Feature.CUSTOM_ATTRIBUTE_FEATURE_TYPE,
+                                     attributeEntry.getValue(), false));
```
aliabbasrizvi: I think `shouldIndex`should be true.

aliabbasrizvi:
Just minor nits now. Mostly looks good.

vraja2:
@aliabbasrizvi Set `shouldIndex` to true, but seems like it's not needed for segmentation.

aliabbasrizvi:
LGTM

PR #23, opened 2016-10-04
Add clientEngine and clientVersion configurability

vraja2:
- Add `withClientEngine` and `withClientVersion` methods for Android to consume.

@optimizely/fullstack-devs


jophde:
LGTM

PR #42, opened 2016-12-14
Add IP anonymization to event payload.

mikeng13:
@optimizely/fullstack-devs

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
+++ b/core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
@@ -79,8 +79,8 @@ public class ProjectConfig {
 
     public ProjectConfig(String accountId, String projectId, String version, String revision, List<Group> groups,
                          List<Experiment> experiments, List<Attribute> attributes, List<EventType> eventType,
-                         List<Audience> audiences, boolean anonymizeIP) {
-        this(accountId, projectId, version, revision, groups, experiments, attributes, eventType, audiences, anonymizeIP,
+                         List<Audience> audiences) {
+        this(accountId, projectId, version, revision, groups, experiments, attributes, eventType, audiences, false,
              null);
     }
 
@@ -239,6 +239,7 @@ public class ProjectConfig {
                 ", projectId='" + projectId + '\'' +
                 ", revision='" + revision + '\'' +
                 ", version='" + version + '\'' +
+                ", anonymizeIP='" + anonymizeIP + '\'' +
                 ", groups=" + groups +
                 ", experiments=" + experiments +
                 ", attributes=" + attributes +
--- a/core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
+++ b/core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
@@ -58,6 +58,7 @@ public class ProjectConfig {
     private final String projectId;
     private final String revision;
     private final String version;
+    private final boolean anonymizeIP;
     private final List<Group> groups;
     private final List<Experiment> experiments;
     private final List<Attribute> attributes;
@@ -78,19 +79,20 @@ public class ProjectConfig {
 
     public ProjectConfig(String accountId, String projectId, String version, String revision, List<Group> groups,
                          List<Experiment> experiments, List<Attribute> attributes, List<EventType> eventType,
-                         List<Audience> audiences) {
-        this(accountId, projectId, version, revision, groups, experiments, attributes, eventType, audiences,
+                         List<Audience> audiences, boolean anonymizeIP) {
+        this(accountId, projectId, version, revision, groups, experiments, attributes, eventType, audiences, anonymizeIP,
              null);
     }
 
     public ProjectConfig(String accountId, String projectId, String version, String revision, List<Group> groups,
                          List<Experiment> experiments, List<Attribute> attributes, List<EventType> eventType,
-                         List<Audience> audiences, List<LiveVariable> liveVariables) {
+                         List<Audience> audiences, boolean anonymizeIP, List<LiveVariable> liveVariables) {
 
         this.accountId = accountId;
         this.projectId = projectId;
         this.version = version;
         this.revision = revision;
+        this.anonymizeIP = anonymizeIP;
 
         this.groups = Collections.unmodifiableList(groups);
         List<Experiment> allExperiments = new ArrayList<Experiment>();
@@ -149,6 +151,10 @@ public class ProjectConfig {
 
     public String getRevision() {
         return revision;
+    }
+
+    public boolean getAnonymizeIP() {
+        return anonymizeIP;
     }
 
     public List<Group> getGroups() {
--- a/core-api/src/main/java/com/optimizely/ab/event/internal/payload/Conversion.java
+++ b/core-api/src/main/java/com/optimizely/ab/event/internal/payload/Conversion.java
@@ -31,12 +31,13 @@ public class Conversion extends Event {
     private List<EventMetric> eventMetrics;
     private List<Feature> eventFeatures;
     private boolean isGlobalHoldback;
+    private boolean anonymizeIP;
 
     public Conversion() { }
 
     public Conversion(String visitorId, long timestamp, String projectId, String accountId, List<Feature> userFeatures,
                       List<LayerState> layerStates, String eventEntityId, String eventName,
-                      List<EventMetric> eventMetrics, List<Feature> eventFeatures, boolean isGlobalHoldback) {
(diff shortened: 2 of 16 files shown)


core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
```diff
@@ -58,6 +58,7 @@ public String toString() {
     private final String projectId;
     private final String revision;
     private final String version;
+    private final boolean anonymizeIP;
```
vraja2: just for consistency, can you modify `toString` to include `anonymizeIP` as well?

core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
```diff
@@ -78,19 +79,20 @@ public String toString() {
 
     public ProjectConfig(String accountId, String projectId, String version, String revision, List<Group> groups,
                          List<Experiment> experiments, List<Attribute> attributes, List<EventType> eventType,
-                         List<Audience> audiences) {
-        this(accountId, projectId, version, revision, groups, experiments, attributes, eventType, audiences,
+                         List<Audience> audiences, boolean anonymizeIP) {
```
vraja2: we only need to add `anonymizeIP` to one `ProjectConfig` constructor. You can remove it here and call the other constructor with `false` as the default value.

core-api/src/test/java/com/optimizely/ab/config/ProjectConfigTestUtils.java
```diff
@@ -139,7 +139,7 @@ private static ProjectConfig generateValidProjectConfigV1() {
                                                  Collections.<TrafficAllocation>emptyList());
         List<Group> groups = asList(randomPolicyGroup, overlappingPolicyGroup);
 
-        return new ProjectConfig("789", "1234", "1", "42", groups, experiments, attributes, events, audiences);
+        return new ProjectConfig("789", "1234", "1", "42", groups, experiments, attributes, events, audiences, true);
```
vraja2: don't need to modify this if you change the constructor

core-api/src/test/java/com/optimizely/ab/config/ProjectConfigTestUtils.java
```diff
@@ -176,7 +176,7 @@ private static ProjectConfig generateNoAudienceProjectConfigV1() {
                                         new EventType("099", "clicked_purchase", multipleExperimentIds));
 
         return new ProjectConfig("789", "1234", "1", "42", Collections.<Group>emptyList(), experiments, attributes,
-                                 events, Collections.<Audience>emptyList());
+                                 events, Collections.<Audience>emptyList(), true);
```
vraja2: don't need to modify

core-api/src/test/java/com/optimizely/ab/config/ProjectConfigTestUtils.java
```diff
@@ -270,7 +270,7 @@ private static ProjectConfig generateValidProjectConfigV2() {
                                                  Collections.<TrafficAllocation>emptyList());
         List<Group> groups = asList(randomPolicyGroup, overlappingPolicyGroup);
 
-        return new ProjectConfig("789", "1234", "2", "42", groups, experiments, attributes, events, audiences);
+        return new ProjectConfig("789", "1234", "2", "42", groups, experiments, attributes, events, audiences, true);
```
vraja2: don't need to modify

core-api/src/test/java/com/optimizely/ab/config/ProjectConfigTestUtils.java
```diff
@@ -307,7 +307,7 @@ private static ProjectConfig generateNoAudienceProjectConfigV2() {
                                         new EventType("099", "clicked_purchase", multipleExperimentIds));
 
         return new ProjectConfig("789", "1234", "2", "42", Collections.<Group>emptyList(), experiments, attributes,
-                                 events, Collections.<Audience>emptyList());
+                                 events, Collections.<Audience>emptyList(), true);
```
vraja2: don't need to modify

core-api/src/test/java/com/optimizely/ab/event/internal/EventBuilderV2Test.java
```diff
@@ -90,6 +90,7 @@ public void createImpressionEvent() throws Exception {
         assertThat(impression.getVisitorId(), is(userId));
         assertThat((double)impression.getTimestamp(), closeTo((double)System.currentTimeMillis(), 60.0));
         assertFalse(impression.getIsGlobalHoldback());
+        assertThat(impression.getAnonymizeIP(), is(projectConfig.getAnonymizeIP()));
```
vraja2: let's make this comparison explicit. assertFalse in this case since it's a v2 datafile.

vraja2: let's make this comparison explicit. assertFalse in this case since it's a v2 datafile. We should also pass in v2 datafile to the eventbuilder as well as a v3 datafile since for v2, `anonymizeI`P should be `false` by default, while v3 will vary

core-api/src/main/java/com/optimizely/ab/event/internal/payload/Conversion.java
```diff
@@ -192,6 +198,7 @@ public String toString() {
                 ", eventMetrics=" + eventMetrics +
                 ", eventFeatures=" + eventFeatures +
                 ", isGlobalHoldback=" + isGlobalHoldback +
+                ", anonymizeIP=" + anonymizeIP +
```
wangjoshuah: should we have constants for all these param values?

core-api/src/test/java/com/optimizely/ab/event/internal/EventBuilderV2Test.java
mikeng13: I don't think testing for that is needed. This tests that events are built correctly and not the fact that IPAnon should be `false` by default as that should be tested elsewhere. We should just make sure that the event value corresponds to what the project config shows. That said, I'll add a test for this case on the `ProjectConfigTest`

core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
mikeng13: Will do

core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
mikeng13: Will do. This will ensure that we always have it initialized to `false` unless explicitly passed in to the constructor.

core-api/src/main/java/com/optimizely/ab/event/internal/payload/Conversion.java
mikeng13: Yeah we should at some point since we share these with `Impression.java`. Just don't think it's worth refactoring for this diff :)

core-api/src/main/java/com/optimizely/ab/event/internal/payload/Impression.java
```diff
@@ -107,6 +109,10 @@ public void setUserFeatures(List<Feature> userFeatures) {
         this.userFeatures = userFeatures;
     }
 
+    public boolean getAnonymizeIP() { return anonymizeIP; }
```
vraja2: nit: can you format this to be on multiple lines?

core-api/src/main/java/com/optimizely/ab/event/internal/payload/Impression.java
```diff
@@ -107,6 +109,10 @@ public void setUserFeatures(List<Feature> userFeatures) {
...
+
+    public void setAnonymizeIP(boolean anonymizeIP) { this.anonymizeIP = anonymizeIP; }
```
vraja2: nit: can you format this to be on multiple lines?

core-api/src/test/java/com/optimizely/ab/config/ProjectConfigTest.java
```diff
@@ -30,8 +30,7 @@
 
 import static java.util.Arrays.asList;
 import static org.hamcrest.CoreMatchers.is;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertThat;
+import static org.junit.Assert.*;
```
vraja2: let's keep explicit imports instead of wildcard imports

core-api/src/test/java/com/optimizely/ab/event/internal/EventBuilderV2Test.java
vraja2: thanks! that makes sense

PR #48, opened 2016-12-22
Add generated BuildConfig class containing version info instead of using a resource file

vraja2:
@optimizely/fullstack-devs


mikeng13:
LGTM

PR #51, opened 2017-01-03
Allow for builds to be triggered on tag creation

vraja2:
@optimizely/fullstack-devs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,3 +20,4 @@ branches:
   only:
   - master
   - devel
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -20,3 +20,4 @@ branches:
   only:
   - master
   - devel
+  - /^\d+\.\d+\.\d+(-SNAPSHOT)?$/ # ensure builds are run on tags which have versions as branch names
```
wangjoshuah: "Trigger builds on tags which are named '<number>.<number>.<number>(-optional)' to ship the SDK."

PR #65, opened 2017-01-27
Elliot/custom builder url

elliotykim:
This change to the Java-SDK Is to support a change in the Android-SDK that allows developers to customize the event endpoints: https://optimizely.atlassian.net/browse/OASIS-809

### DIFF
--- a/core-api/src/test/java/com/optimizely/ab/event/internal/EventBuilderV2Test.java
+++ b/core-api/src/test/java/com/optimizely/ab/event/internal/EventBuilderV2Test.java
@@ -82,7 +82,7 @@ public class EventBuilderV2Test {
                                                                  userId, attributeMap);
 
         // verify that request endpoint is correct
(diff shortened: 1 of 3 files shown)


core-api/src/test/java/com/optimizely/ab/event/internal/EventBuilderV2Test.java
```diff
@@ -147,6 +147,28 @@ public void createImpressionEventCustomClientEngineClientVersion() throws Except
     }
 
     /**
+     * Verify that supplying {@link EventBuilderV2} with a custom impression endpoint and conversion endpoint results
+     * in impression events being sent to the overridden URL.
+     */
+    @Test
+    public void createImpressionEventCustomEndpoints() throws Exception {
```
aliabbasrizvi: It is both impression and conversion.

elliotykim: @aliabbasrizvi the conversion event is tested below in a separate test case

PR #73, opened 2017-02-15
Add Android TV SDK client engine type to Event

onufryk:
Summary: Set client engine to android tv if SDK is used on AndroidTV


vraja2:
@onufryk nice, just make sure to squash and merge into devel

PR #76, opened 2017-02-27
Catch any exception occured during parsing project config

onufryk:
CCing @optimizely/fullstack-devs


core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -498,7 +498,13 @@ private void track(@Nonnull String eventName,
      * @return a {@link ProjectConfig} instance given a json string
      */
     private static ProjectConfig getProjectConfig(String datafile) throws ConfigParseException {
-        return DefaultConfigParser.getInstance().parseProjectConfig(datafile);
+        ProjectConfig projectConfig;
+        try {
+            projectConfig = DefaultConfigParser.getInstance().parseProjectConfig(datafile);
```
elliotykim: this line could still be `return DefaultConfigParser.getInstance().parseProjectConfig(datafile);`

then we don't need to define an extra variable

PR #77, opened 2017-02-27
Test null, empty and malformed data corner cases

onufryk:


core-api/src/test/java/com/optimizely/ab/OptimizelyBuilderTest.java
```diff
@@ -180,6 +180,7 @@ public void builderThrowsConfigParseExceptionForEmptyDatafile() throws Exception
     @Test
     public void builderThrowsConfigParseExceptionForInvalidDatafile() throws Exception {
         thrown.expect(ConfigParseException.class);
-        Optimizely.builder("{invalidDatafile}", mockEventHandler).build();
+        String invalidJson = "{invalidDatafile}";
```
vraja2: since this was just for rebuilding, can we change this back?

PR #79, opened 2017-02-28
`filterAttributes` returns an empty if `attributes` parameter is null

elliotykim:


elliotykim:
@optimizely/fullstack-devs

onufryk:
Don't forget to squash before merge

PR #81, opened 2017-03-09
Use `@BeforeClass` to do static set up.

wangjoshuah:
The datafiles and configs need to only be set up once per test.
Takes about 1 second of each test case run time.


wangjoshuah:
@optimizely/fullstack-devs

vraja2:
make sure to squash and merge. thanks!

PR #90, opened 2017-04-26
remove license check in gradle build

wangjoshuah:


wangjoshuah:
@optimizely/fullstack-devs

PR #129, opened 2017-07-15
Add support for numeric metrics

vraja2:
@optimizely/fullstack-devs

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilderV2.java
+++ b/core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilderV2.java
@@ -112,10 +112,16 @@ public class EventBuilderV2 extends EventBuilder {
 
         List<LayerState> layerStates = createLayerStates(projectConfig, experimentVariationMap);
 
-        Long eventValue = EventTagUtils.getRevenueValue(eventTags);
-        List<EventMetric> eventMetrics = Collections.emptyList();
-        if (eventValue != null) {
-            eventMetrics = Collections.singletonList(new EventMetric(EventMetric.REVENUE_METRIC_TYPE, eventValue));
+        List<EventMetric> eventMetrics = new ArrayList<EventMetric>();
+
+        Long revenueValue = EventTagUtils.getRevenueValue(eventTags);
+        if (revenueValue != null) {
+            eventMetrics.add(new EventMetric(EventMetric.REVENUE_METRIC_TYPE, revenueValue));
+        }
+
+        Double numericMetricValue = EventTagUtils.getNumericValue(eventTags);
+        if (numericMetricValue != null) {
+            eventMetrics.add(new EventMetric(EventMetric.NUMERIC_METRIC_TYPE, numericMetricValue));
         }
 
         Conversion conversionPayload = new Conversion();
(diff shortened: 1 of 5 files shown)


core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilderV2.java
```diff
@@ -112,10 +112,16 @@ public LogEvent createConversionEvent(@Nonnull ProjectConfig projectConfig,
...
-        if (eventValue != null) {
-            eventMetrics = Collections.singletonList(new EventMetric(EventMetric.REVENUE_METRIC_TYPE, eventValue));
+        List<EventMetric> eventMetrics = new ArrayList<EventMetric>();
+
+        Long revenueValue = EventTagUtils.getRevenueValue(eventTags);
+        if (revenueValue != null) {
+            eventMetrics.add(new EventMetric(EventMetric.REVENUE_METRIC_TYPE, revenueValue));
+        }
+
+        Double numericMetricValue = EventTagUtils.getNumericValue(eventTags);
```
wangjoshuah: @vraja2 can there only be one numeric metric value?

core-api/src/main/java/com/optimizely/ab/event/internal/payload/EventMetric.java
```diff
@@ -38,30 +39,29 @@ public void setName(String name) {
...
+    public void setValue(Number value) {
         this.value = value;
     }
 
-
     @Override
-    public boolean equals(Object other) {
-        if (!(other instanceof EventMetric))
-            return false;
+    public boolean equals(Object o) {
```
wangjoshuah: please don't name variables single characters

core-api/src/main/java/com/optimizely/ab/internal/EventTagUtils.java
```diff
@@ -47,4 +47,22 @@ public static Long getRevenueValue(@Nonnull Map<String, ?> eventTags) {
...
     }
+
+    /**
+     * Fetch the numeric metric value from event tags. "value" is a reserved keyword.
+     */
+    public static Double getNumericValue(@Nonnull Map<String, ?> eventTags) {
+        Double eventValue = null;
+        if (eventTags.containsKey(ReservedEventKey.VALUE.toString())) {
+            Object rawValue = eventTags.get(ReservedEventKey.VALUE.toString());
+            if (rawValue instanceof Double) {
```
loganlinn: It's reasonable to let this be anything that implements `Number` IMO. I think this will lead to confusing behavior

wangjoshuah: +1

core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilderV2.java
vraja2: yes, it will correspond to the event being tracked.

core-api/src/main/java/com/optimizely/ab/event/internal/payload/EventMetric.java
vraja2: intellij auto-generated, will change to `obj`

core-api/src/main/java/com/optimizely/ab/internal/EventTagUtils.java
vraja2: makes sense, the opposite case is the one that we should be wary of (truncating floating-point values)

wangjoshuah:
@vraja2 will add docs to explain why they should send us floats or doubles instead of int/long

PR #131, opened 2017-08-07
internal feature variable accessor support

wangjoshuah:
Create Feature Variable Accessor with type support
Create internal Decision Service method for features


wangjoshuah:
@optimizely/fullstack-devs

PR #133, opened 2017-08-16
feature variable accessor apis

wangjoshuah:
* link feature variable accessors to feature variable by value and type
* add unit tests for getFeatureVariable calling through to internals
* add unit tests to catch parsing error


wangjoshuah:
@optimizely/fullstack-devs

PR #142, opened 2017-09-12
ignore experiment status when bucketing rollout rule experiments

wangjoshuah:
we do not need to check experiment status since the datafile will not have inactive experiments


wangjoshuah:
@thomaszurkan-optimizely The rollout rule experiment will be filtered out if not active by the datafile.

PR #145, opened 2017-09-25
use trusty distribution

wangjoshuah:
Upgrade travis-ci `travis.yml` to use `trusty` distribution since that is their new standard.
The gradle version of their linux distribution is gradle `4.0.1` so I upgraded our gradle wrapper to that.
That required we upgrade our nebula dependency to `3.2.0`.
Drop support for JDK 7 until travis-ci can fix that.


core-api/build.gradle
```diff
@@ -2,14 +2,14 @@ dependencies {
...
-    provided group: 'com.google.code.findbugs', name: 'jsr305', version: findbugsVersion
+    compile group: 'com.google.code.findbugs', name: 'annotations', version: findbugsVersion
+    compile group: 'com.google.code.findbugs', name: 'jsr305', version: findbugsVersion
 
     // an assortment of json parsers
-    provided group: 'com.google.code.gson', name: 'gson', version: gsonVersion, optional
-    provided group: 'org.json', name: 'json', version: jsonVersion, optional
-    provided group: 'com.googlecode.json-simple', name: 'json-simple', version: jsonSimpleVersion, optional
-    provided group: 'com.fasterxml.jackson.core', name: 'jackson-databind', version: jacksonVersion, optional
+    compile group: 'com.google.code.gson', name: 'gson', version: gsonVersion, optional
```
vraja2: Changing these to `compile` will result in all of these JSON libraries being packaged with the SDK which will increase bloat. `provided` was used to give that dependency flexibility to the consumer.

.travis.yml
```diff
@@ -1,13 +1,14 @@
 language: java
-dist: precise
+dist: trusty
+sudo: required
```
vraja2: any reason we need root privileges?

.travis.yml
```diff
@@ -1,13 +1,14 @@
...
 jdk:
-  - openjdk7
```
vraja2: I think `openjdk7` should still be supported.

wangjoshuah:
@vraja2 travis-ci.org is ending support for `dist: precise`. We are migrating to `dist: trusty` do ensure support. https://blog.travis-ci.com/2017-08-31-trusty-as-default-status

The `trusty` distribution uses `gradle 4.0.1` which is incompatible with our current version of `nebula`. We will use `nebula 3.2.0` to allow us to continue testing all parsers. https://github.com/nebula-plugins/gradle-extra-configurations-plugin

PR #152, opened 2017-10-25
patch user attribute evaluation

wangjoshuah:


wangjoshuah:
build

wangjoshuah:
build

nchilada:
@wangjoshuah @aliabbasrizvi do you remember why we made this change? AFAIK the datafile's audience conditions should always include non-null `value`s.

PR #162, opened 2018-01-05
2.0.0 beta3 bump

wangjoshuah:
* Bump version to 2.0.0-beta
* Update CHANGELOG.md


gradle.properties
```diff
@@ -1,5 +1,5 @@
 # Maven version
-version = 2.0.0-SNAPSHOT
+version = 2.0.0-BETA
```
vraja2: `SNAPSHOT` is conventional so don't need to replace it. The tag specified when drafting a release through Github will override this version so you can define it as `2.0.0-beta3` there (looks like it needs to be lowercase https://github.com/optimizely/java-sdk/blob/master/.travis.yml#L24, probably should be changed to caps in the future)

PR #166, opened 2018-01-31
rename ActivateNotification to ActivateNotificationListener same with…

thomaszurkan-optimizely:
… TrackNotificationListener


core-api/src/test/java/com/optimizely/ab/notification/NotificationCenterTest.java
```diff
@@ -1,30 +1,34 @@
 package com.optimizely.ab.notification;
 
 import ch.qos.logback.classic.Level;
+import com.optimizely.ab.config.Experiment;
```
wangjoshuah: seems like you have some unused imports. make sure to clear them

PR #169, opened 2018-02-24
integrate coveralls

wangjoshuah:


wangjoshuah:
Integrate multi-project build of coveralls with jacoco from guide here
https://github.com/kt3k/coveralls-gradle-plugin#use-with-mutli-project-build

PR #175, opened 2018-03-22
refactor so that any java 8 or later can use lambdas for Activate and…

thomaszurkan-optimizely:
… Track listeners.


core-api/src/main/java/com/optimizely/ab/notification/NotificationCenter.java
```diff
@@ -16,9 +16,13 @@
  */
 package com.optimizely.ab.notification;
 
+import com.optimizely.ab.config.Experiment;
```
mikeng13: Can you change the license header to say 2017-2018. Let's make sure to do this for the files we touch in the PR

core-api/src/main/java/com/optimizely/ab/notification/NotificationCenter.java
```diff
@@ -78,6 +82,43 @@ public NotificationCenter() {
     // we used a list so that notification order can mean something.
     private Map<NotificationType, ArrayList<NotificationHolder>> notificationsListeners =new HashMap<NotificationType, ArrayList<NotificationHolder>>();
 
+    /**
+     * Convenience method to support lambdas as callbacks in later version of Java (8+).
+     * @param activateNotificationListenerInterface
+     * @return greater than zero if added.
```
mikeng13: I wonder if we will run into concurrency/threading issues with incrementing this static variable. What if we end up with the same ID for different listeners?

core-api/src/main/java/com/optimizely/ab/notification/NotificationListener.java
```diff
@@ -26,53 +26,17 @@
...
-    @Deprecated
-    public void onExperimentActivated(@Nonnull Experiment experiment,
-                                      @Nonnull String userId,
-                                      @Nonnull Map<String, String> attributes,
-                                      @Nonnull Variation variation) {
-    }
+public interface NotificationListener {
 
     /**
      * This is the new method of notification.  Implementation classes such as {@link ActivateNotificationListener}
```
mikeng13: just a nitpick: No need to say this method is new. I think we can handle that in the changelogs

core-api/src/main/java/com/optimizely/ab/notification/NotificationListener.java
```diff
@@ -26,53 +26,17 @@
...
      * will implement this call and provide another method with the correct parameters
      * Notify called when a notification is triggered via the {@link com.optimizely.ab.notification.NotificationCenter}
      * @param args - variable argument list based on the type of notification.
      */
-    public abstract void notify(Object... args);
+    public void notify(Object... args);
```
mikeng13: instead of having variables args and attempting to parse them in the implementation, why not have a static notification object? I think that would be way easier to read in code

core-api/src/main/java/com/optimizely/ab/notification/NotificationCenter.java
thomaszurkan-optimizely: not static.  you are looking at the holder class def and not the notification id.

wangjoshuah:
can you try this out in a testapp to verify?

thomaszurkan-optimizely:
build

PR #178, opened 2018-04-12
Bump to 2.0 for General Availability.

mikeng13:


CHANGELOG.md
```diff
@@ -1,5 +1,41 @@
...
+* The `track` API with revenue value as a stand-alone parameter has been removed. The revenue value should be passed in as an entry of the event tags map. The key for the revenue tag is `revenue` and will be treated by Optimizely as the key for analyzing revenue data in results.
+```
+Map<String, Object> eventTags = new HashMap<String, Object>();
+
+// reserved "revenue" tag
+eventTags.put("revenue", 6432);
+
+optimizelyClient.track("event_key", "user_id", userAttributes, eventTags);
+```
+
```
thomaszurkan-optimizely: Can we add the following to Breaking Changes?:

We have removed deprecated classes with the NotificationBroadcaster in favor of the new API with the NotificationCenter. We have streamlined the API so that it is easily usable with Java Lambdas in Java 1.8+. We have also added some convenience methods to add these listeners. Finally, some of the API names have changed slightly (e.g. clearAllNotifications() is now clearAllNotificationListeners())

PR #179, opened 2018-04-12
[WIP] Refactor experiment and events getters

wangjoshuah:
* Remove throws declarations on methods since we never actually throw. 
* Change docs to explain that we send it to the error handler instead

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
+++ b/core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
@@ -17,11 +17,16 @@
 package com.optimizely.ab.config;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.optimizely.ab.UnknownExperimentException;
 import com.optimizely.ab.config.audience.Audience;
 import com.optimizely.ab.config.audience.Condition;
+import com.optimizely.ab.error.ErrorHandler;
+import com.optimizely.ab.error.NoOpErrorHandler;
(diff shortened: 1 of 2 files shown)


core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
```diff
@@ -210,6 +216,62 @@ public ProjectConfig(String accountId,
...
+     */
+    @CheckForNull
+    public Experiment getExperimentForKey(@Nonnull String experimentKey,
+                                          @Nonnull ErrorHandler errorHandler) {
+
+        Experiment experiment =
+                getExperimentKeyMapping()
+                        .get(experimentKey);
+
+        // if the given experiment key isn't present in the config, log and potentially throw an exception
```
mikeng13: Revise comment on throwing error

core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
```diff
@@ -210,6 +216,62 @@ public ProjectConfig(String accountId,
...
+     * @param errorHandler the error handler to send exceptions to
+     * @return the event type for the given event name
+     */
+    public @CheckForNull EventType getEventTypeForName(String eventName, ErrorHandler errorHandler) {
+
+        EventType eventType =
+                getEventNameMapping()
+                        .get(eventName);
+
+        // if the given event name isn't present in the config, log and potentially throw an exception
```
mikeng13: No exceptions anymore!

PR #180, opened 2018-04-19
getEnabledFeature, Activate, getVariation and track Null check fixes

mnoman09:
* Added test Case of Non null user id in get enabled Feature

* Added Null check of experiment in user activate

* added null check in user id and in event key of track

* Added null check on Experiment key of getVariation and added its unitTest

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/Optimizely.java
+++ b/core-api/src/main/java/com/optimizely/ab/Optimizely.java
@@ -122,6 +122,11 @@ public class Optimizely {
                        @Nonnull String userId,
                        @Nonnull Map<String, String> attributes) throws UnknownExperimentException {
 
+        if (experimentKey == null) {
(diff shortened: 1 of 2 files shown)


core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -584,7 +604,7 @@ else if (userId == null) {
      * @return List of the feature keys that are enabled for the user if the userId is empty it will
      * return Empty List.
      */
-    public List<String> getEnabledFeatures(@Nonnull String userId,@Nonnull Map<String, String> attributes) {
+    public List<String> getEnabledFeatures(@Nonnull String userId, @Nonnull Map<String, String> attributes) throws IllegalArgumentException {
```
thomaszurkan-optimizely: I don't think this should have a throws.  Any of the methods could throw an illegal argument exception because of the non null annotations.

PR #184, opened 2018-06-05
Added BotFiltering

mnoman09:
* Added BotFiltering
* Added unit tests

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
+++ b/core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
@@ -1,6 +1,6 @@
 /**
  *
- *    Copyright 2016-2017, Optimizely and contributors
+ *    Copyright 2016-2018, Optimizely and contributors
  *
  *    Licensed under the Apache License, Version 2.0 (the "License");
  *    you may not use this file except in compliance with the License.
@@ -24,8 +24,10 @@ import com.optimizely.ab.config.audience.Condition;
 import com.optimizely.ab.error.ErrorHandler;
 import com.optimizely.ab.error.NoOpErrorHandler;
 import com.optimizely.ab.error.RaiseExceptionErrorHandler;
+import com.optimizely.ab.internal.ReservedAttributeKey;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.w3c.dom.Attr;
 
 import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
@@ -37,6 +39,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.logging.ErrorManager;
 
 /**
  * Represents the Optimizely Project configuration.
@@ -73,6 +76,7 @@ public class ProjectConfig {
     private final String revision;
     private final String version;
     private final boolean anonymizeIP;
+    private final Boolean botFiltering;
     private final List<Attribute> attributes;
     private final List<Audience> audiences;
     private final List<EventType> events;
@@ -100,6 +104,8 @@ public class ProjectConfig {
     private final Map<String, Map<String, LiveVariableUsageInstance>> variationToLiveVariableUsageInstanceMapping;
     private final Map<String, Experiment> variationIdToExperimentMapping;
 
+    private final static String RESERVED_ATTRIBUTE_PREFIX = "$opt_";
+
     /**
      *  Forced variations supersede any other mappings.  They are transient and are not persistent or part of
      * the actual datafile. This contains all the forced variations
@@ -123,6 +129,7 @@ public class ProjectConfig {
         this(
                 accountId,
                 anonymizeIP,
+                null,
                 projectId,
                 revision,
                 version,
@@ -140,6 +147,7 @@ public class ProjectConfig {
     // v4 constructor
     public ProjectConfig(String accountId,
                          boolean anonymizeIP,
+                         Boolean botFiltering,
                          String projectId,
                          String revision,
                          String version,
@@ -157,6 +165,7 @@ public class ProjectConfig {
         this.version = version;
         this.revision = revision;
         this.anonymizeIP = anonymizeIP;
+        this.botFiltering = botFiltering;
 
         this.attributes = Collections.unmodifiableList(attributes);
         this.audiences = Collections.unmodifiableList(audiences);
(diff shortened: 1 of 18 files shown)


core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
```diff
@@ -140,6 +145,7 @@ public ProjectConfig(String accountId, String projectId, String version, String
     // v4 constructor
     public ProjectConfig(String accountId,
                          boolean anonymizeIP,
+                         Boolean botFiltering,
```
aliabbasrizvi: Why is this `Boolean` and not `boolean` like `anonymizeIP` or vice versa?

core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
```diff
@@ -285,6 +292,35 @@ public Experiment getExperimentForKey(@Nonnull String experimentKey,
         return groupExperiments;
     }
 
+    /**
+     * Checks is attributeKey is reserved or not and is it exist in attributeKeyMapping
```
aliabbasrizvi: nit. and `...if it exists..` and not `...is it exist...`

core-api/src/main/java/com/optimizely/ab/internal/ReservedAttributeKey.java
```diff
@@ -0,0 +1,34 @@
...
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+package com.optimizely.ab.internal;
+
+public enum ReservedAttributeKey {
```
aliabbasrizvi: Perhaps call it `ControlAttribute` like other SDKs.

core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
mnoman09: Because  Boolean is nullable and boolean is not.

thomaszurkan-optimizely: The interesting thing about what you are saying here is that anonymizeIP is also optional based on datafile type (for instance v2 does not contain anonymizeIP).

thomaszurkan-optimizely:
build

core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
mnoman09: anonymizeIp default value is false and we are not checking in our sdk if its null it's just getting passed false if not set, but for botfiltering we have three values which are True, False and Null, if botFiltering value is null than we don't add its attribute in list.

mnoman09:
build

core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
aliabbasrizvi: The same holds true for `anonymizeIP` as it can be null in an older datafile.

core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
```diff
@@ -285,6 +292,35 @@ public Experiment getExperimentForKey(@Nonnull String experimentKey,
...
 
+    /**
+     * Checks is attributeKey is reserved or not and if it exist in attributeKeyMapping
+     * @param attributeKey
+     * @return AttributeId corresponding to AttributeKeyMapping, AttributeKey when it's a reserved attribute and
+     * null when attributeKey is equal to BOT_FILTERING_ATTRIBUTE key.
+     */
+    public String getAttributeId(ProjectConfig projectConfig, String attributeKey) {
+        String attributeIdOrKey = null;
+        if (!attributeKey.equals(ControlAttribute.BOT_FILTERING_ATTRIBUTE.toString())) {
```
aliabbasrizvi: This check is unnecessary. I know why you put it here, but it is an overkill.

core-api/src/test/java/com/optimizely/ab/event/internal/EventBuilderTest.java
```diff
@@ -85,25 +98,93 @@ public EventBuilderTest(int datafileVersion,
...
+        com.optimizely.ab.event.internal.payload.Attribute feature = new com.optimizely.ab.event.internal.payload.Attribute(attribute.getId(),
+                attribute.getKey(), com.optimizely.ab.event.internal.payload.Attribute.CUSTOM_ATTRIBUTE_TYPE,
+                "value");
+        com.optimizely.ab.event.internal.payload.Attribute userAgentFeature = new com.optimizely.ab.event.internal.payload.Attribute(
+                ControlAttribute.USER_AGENT_ATTRIBUTE.toString(),
+                ControlAttribute.USER_AGENT_ATTRIBUTE.toString(),
+                com.optimizely.ab.event.internal.payload.Attribute.CUSTOM_ATTRIBUTE_TYPE,
+                "Chrome");
+
+        com.optimizely.ab.event.internal.payload.Attribute BotFilteringFeature = new com.optimizely.ab.event.internal.payload.Attribute(
```
aliabbasrizvi: I am confused by the styling here. Shouldn't this be `botFilteringFeature`?

aliabbasrizvi:
I think I know the problem now. You need to pass in bot filtering in attributes as boolean and not as string. That is a requirement of the API.

core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
```diff
@@ -285,6 +292,30 @@ public Experiment getExperimentForKey(@Nonnull String experimentKey,
...
 
+    /**
+     * Checks is attributeKey is reserved or not and if it exist in attributeKeyMapping
+     * @param attributeKey
+     * @return AttributeId corresponding to AttributeKeyMapping, AttributeKey when it's a reserved attribute and
+     * null when attributeKey is equal to BOT_FILTERING_ATTRIBUTE key.
+     */
+    public String getAttributeId(ProjectConfig projectConfig, String attributeKey) {
+        String attributeIdOrKey = null;
+        com.optimizely.ab.config.Attribute attribute = projectConfig.getAttributeKeyMapping().get(attributeKey);
```
aliabbasrizvi: Why the inline import? Why not at the head of the file?

PR #187, opened 2018-06-11
Added SourceClear TravisCI configuration to run once per build

ooola:


ooola:
This change enables the SourceClear scanner to run in a separate build. Note that it may duplicate whatever is done in the standard openjdk8 build. 

This is the most straightforward way I found to enable SourceClear to run only once per pull request.

sourceclear.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+if [ "$RUN_SOURCECLEAR" = "true" ]; then
+    echo "RUN_SOURCECLEAR is set, running sourceclear"
+    curl -sSL https://download.sourceclear.com/ci.sh |  bash
+else
+    echo "RUN_SOURCECLEAR it not set, skipping"
```
aliabbasrizvi: nit. is and not it

sourceclear.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
```
aliabbasrizvi: Not a fan of top level sourceclear.sh file. Please create a scripts folder and throw it in there.

wangjoshuah: +1. Same model is used in other SDKs see https://github.com/optimizely/objective-c-sdk/tree/master/Scripts

ooola:
This was just a PoC. Closing for now.

PR #197, opened 2018-06-29
SDK Crash Proofing: Added try/catch blocks at entry points

zashraf1985:
I have incorporated all the feedback that i received as part of similar activity in Javascript SDK.
1) Added try/catch blocks in all public methods of Optimizely, Builder and NotificationCenter
2) Logging the exception using logger instance inside catch block
3) Some methods throw checked exceptions. catching and re-throwing checked exception to make sure all unit tests pass.

Question/Concern: Referring back to point # 3, In order to make sure all unit tests pass, i have to catch specific exceptions and re-throw them. If i do not do that, some unit tests fail because they are expecting those exceptions. Should we go ahead and modify those unit tests too?

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/Optimizely.java
+++ b/core-api/src/main/java/com/optimizely/ab/Optimizely.java
@@ -115,50 +115,71 @@ public class Optimizely {
     public @Nullable
     Variation activate(@Nonnull String experimentKey,
                        @Nonnull String userId) throws UnknownExperimentException {
-        return activate(experimentKey, userId, Collections.<String, String>emptyMap());
+        try {
+            return activate(experimentKey, userId, Collections.<String, String>emptyMap());
+        } catch (UnknownExperimentException e) {
+            throw e;
+        } catch (Exception e) {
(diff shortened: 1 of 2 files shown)


core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -115,76 +115,101 @@ void initialize() {
     public @Nullable
     Variation activate(@Nonnull String experimentKey,
                        @Nonnull String userId) throws UnknownExperimentException {
-        return activate(experimentKey, userId, Collections.<String, String>emptyMap());
+        try {
+            return activate(experimentKey, userId, Collections.<String, String>emptyMap());
+        } catch (UnknownExperimentException e) {
+            throw e;
+        } catch (Exception e) {
```
wangjoshuah: Send the unknown exception to the error handler. We shouldn't hide run time or other unchecked exceptions that might affect the customer's application. They need to have the opportunity of handling the exception in addition to having an error logged.

zashraf1985: Done

PR #198, opened 2018-07-06
prime the serializer in a helper thread to cache reflection calls mak…

thomaszurkan-optimizely:
…ing subsequent calls to create and serialize EventBatch must faster.

This reduces activate calls on Android from 100 ms to 15 ms.

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilder.java
+++ b/core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilder.java
@@ -116,6 +116,9 @@ public class EventBuilder {
                                     @Nonnull String userId,
                                     @Nonnull Map<String, String> attributes) {
         try {
+            // guard 
+            if (primedEvent == null) return null;
+
             EventBatch eventBatch = primedEvent.get();
             primedEvent = null;
             Decision decision = new Decision(activatedExperiment.getLayerId(), activatedExperiment.getId(),
--- a/core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilder.java
+++ b/core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilder.java
@@ -41,6 +41,8 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 public class EventBuilder {
     private static final Logger logger = LoggerFactory.getLogger(EventBuilder.class);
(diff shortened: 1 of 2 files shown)


core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilder.java
```diff
@@ -61,8 +63,32 @@ public EventBuilder(EventBatch.ClientEngine clientEngine, String clientVersion)
...
+                primeSerializer();
+            }
+        };
+
+        ExecutorService executor = Executors.newCachedThreadPool();
+        executor.submit(r);
+
     }
 
+    private void primeSerializer() {
```
mikeng13: Can you add a comment explaining why this is necessary?

core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilder.java
```diff
@@ -61,8 +63,32 @@ public EventBuilder(EventBatch.ClientEngine clientEngine, String clientVersion)
...
+        Decision decision = new Decision("", "",
```
mikeng13: I'd like to be on the safer side and put this inside a `try/catch`. Worst that can happen then is that the priming fails and the very first API call that sends events will take longer

thomaszurkan-optimizely: Because it is in its own Runnable thread, worst case is thread dies.

jophde:
Do we have any benchmarks around how expensive of call this is caching?

thomaszurkan-optimizely:
@jophde I wrapped some times around things and discovered that the creation of the EventBatch and its associated objects made up about 60% of the time while the serializer took up another 40% of a total somewhere over 100 ms.  It now sub 100 ms.

core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilder.java
```diff
@@ -61,21 +67,77 @@ public EventBuilder(EventBatch.ClientEngine clientEngine, String clientVersion)
...
     }
 
+    private void waitForPrimedEvent() {
+        if (primedEvent != null) {
+            logger.debug("had to wait for prime");
+            try {
+                primedEvent.get();
+            }
+            catch (Exception e) {
+                logger.error("Problem getting lock ", e);
```
mikeng13: Can you give a bit context in this message so the user will have an easier time figuring what's wrong?

core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilder.java
```diff
@@ -61,21 +67,77 @@ public EventBuilder(EventBatch.ClientEngine clientEngine, String clientVersion)
         this.clientEngine = clientEngine;
         this.clientVersion = clientVersion;
         this.serializer = DefaultJsonSerializer.getInstance();
+
+        Callable<EventBatch> callable = new Callable<EventBatch>() {
+            @Override
+            public EventBatch call() throws Exception {
+                EventBatch eventBatch = null;
+                try {
+                    eventBatch = primeSerializer();
```
mikeng13: Should this try to return?

core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilder.java
```diff
@@ -61,21 +67,77 @@ public EventBuilder(EventBatch.ClientEngine clientEngine, String clientVersion)
...
+                "", false);
+        Event impressionEvent = new Event(System.currentTimeMillis(),UUID.randomUUID().toString(), "",
+                ACTIVATE_EVENT_KEY, null, null, null, ACTIVATE_EVENT_KEY, null);
+
+        Snapshot snapshot = new Snapshot(Arrays.asList(decision), Arrays.asList(impressionEvent));
+        Visitor visitor = new Visitor("", null, new ArrayList<Attribute>(), Arrays.asList(snapshot));
+        List<Visitor> visitors = Arrays.asList(visitor);
+        EventBatch eventBatch = new EventBatch(clientEngine.getClientEngineValue(), clientVersion, "", visitors, true, "", "");
+        String payload = this.serializer.serialize(eventBatch);
+        logger.debug("JSON Serializer primed for payload {}", payload);
```
mikeng13: Nit: ... for event payload

core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilder.java
```diff
@@ -61,21 +67,77 @@ public EventBuilder(EventBatch.ClientEngine clientEngine, String clientVersion)
...
+
+        return  eventBatch;
     }
 
+    private void waitForPrimedEvent() {
+        if (primedEvent != null) {
+            logger.debug("had to wait for prime");
```
mikeng13: This message could be a bit more descriptive

core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilder.java
```diff
@@ -61,21 +67,77 @@ public EventBuilder(EventBatch.ClientEngine clientEngine, String clientVersion)
...
         this.serializer = DefaultJsonSerializer.getInstance();
+
+        Callable<EventBatch> callable = new Callable<EventBatch>() {
+            @Override
+            public EventBatch call() throws Exception {
+                EventBatch eventBatch = null;
+                try {
+                    eventBatch = primeSerializer();
+                } catch (Exception e) {
+                    logger.error("Problem priming the JsonSerializer ", e);
```
mikeng13: This error message is too generic. Let's give the user something more specific

thomaszurkan-optimizely:
I want to show one other option in a different pr

PR #199, opened 2018-07-11
this fix brings down activate to 5 ms.  We should do it for track as well …

thomaszurkan-optimizely:
…as well if there is no objection

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilder.java
+++ b/core-api/src/main/java/com/optimizely/ab/event/internal/EventBuilder.java
@@ -46,7 +46,7 @@ public class EventBuilder {
     private static final Logger logger = LoggerFactory.getLogger(EventBuilder.class);
     static final String EVENT_ENDPOINT = "https://logx.optimizely.com/v1/events";
     static final String ACTIVATE_EVENT_KEY = "campaign_activated";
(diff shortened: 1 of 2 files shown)


wangjoshuah:
Agree with Mike that this is very hard to read and potentially brittle if we change the event template in the future.

If you really want to shave time off of event serialization, I'd suggest using something like the Java `String formatter` to do the replacing instead of individual `String.replace()` methods. See reference here https://docs.oracle.com/javase/8/docs/api/java/util/Formatter.html

Also, if you are making performance improvements, I would recommend also adding performance testing to prove the improvement as well as to prevent regressions.

thomaszurkan-optimizely:
@wangjoshuah thanks for the feedback.  I did update the template with comments and indenting.  I feel that the template is much more obvious than using a formatter.  Also, we are continuing to use the json serializers and objects for unit testing.  That way we should catch any change that was made on one side and not the other.

mikeng13:
Please get the travis build passing before you merge!

thomaszurkan-optimizely:
I have another commit where i do the conversion event as well.  in order to do that, i had to make it more flexible.  ideally, we should have these sting templates at each class level (i.e. Event.TEMPLATE, EventBatch.TEMPLATE) and then put them together to either make a single event batch or multiple.  :)   but, this may be "good enough" for now. @mikeng13 ?

mikeng13:
Yeah is should be good enough

thomaszurkan-optimizely:
build

thomaszurkan-optimizely:
build

aliabbasrizvi:
I am concerned about the readability here. I would like to explore if there is a neater way to achieve this.

aliabbasrizvi:
@mikeborsuk-optimizely @loganlinn @mikecdavis can you help review this?

aliabbasrizvi:
Also @vraja2

aliabbasrizvi:
Thanks for this breakdown @mikecdavis

aliabbasrizvi:
@thomaszurkan-optimizely we can try and run the benchmarking tests against this branch as well: https://github.com/optimizely/java-sdk/tree/ali/benchmark_java

The results will be similar to the analysis @mikecdavis did.

PR #201, opened 2018-08-01
Move serialization to LogEvent.getBody().

mikecdavis:
This PR moves the serialization from POJO to JSON into `LogEvent#getBody()`. This allows the serialization of the event to be offloaded to with the event handler when the HTTP request is built. This should be 100% backwards compatible with any existing `EventHandler` implementations. Initial benchmarks are much improved over the current master in tight loop simulations.

Stats after 100,000 calls to `activate(...)` in my simulation:

|Metric (us)|PR-199|2.0.0-beta2|2.0.0-SNAPSHOT|THIS PR|
|---|---|---|---|---|
|Min|21|2|3|1||
|Median|29|2|9|6||
|Mean|41|3.8|10.2|6.1||
|75th Pct|40|3|11|7||
|95th Pct|85|7|18|11||
|99th Pct|126|26|43|21||
|999th Pct|713|65|237|232||
|Max|11825|17732|34384|12845||

Note `2.0.0-beta2` was pre-batch endpoint, so isn't a valid comparison. Also performance runs can have some volatility, but the overall trends remain.

I've also wrapped some calls to `logger.debug` in `logger.isDebugEnabled()` to prevent unused log statement from triggering the serialization synchronously.

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/Optimizely.java
+++ b/core-api/src/main/java/com/optimizely/ab/Optimizely.java
@@ -199,9 +199,13 @@ public class Optimizely {
                     userId,
                     filteredAttributes);
             logger.info("Activating user \"{}\" in experiment \"{}\".", userId, experiment.getKey());
-            logger.debug(
-                    "Dispatching impression event to URL {} with params {} and payload \"{}\".",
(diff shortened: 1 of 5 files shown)


aliabbasrizvi:
THANK YOU !! Taking a closer look now.

ceimaj:
@thomaszurkan-optimizely can we get this into the 2.1 Android release?

PR #204, opened 2018-08-10
Include Credits section in README.md.

mikeng13:


mikeng13:
That's a good point, I don't think we need `FindBug` in there. Let me check with @thomaszurkan-optimizely to see how he did it.

PR #206, opened 2018-08-16
update the async event handler to kotlin as an example

thomaszurkan-optimizely:


thomaszurkan-optimizely:
closing for now until we can support java sdk 1.8

PR #207, opened 2018-08-21
feat(api): Accepting all types for attributes values

shihpatrick:
## Summary
As a part of Audience Match Type, I changed all relevant function signatures so that it accepts an `Object` variable rather than just a `String` for functions that accept and uses `attributes`. Any filtering of types would be done later, when evaluating audience conditions. 

## Testing
Passes all current unit tests. Further testing of other types for attribute values (like filtering out types that aren't `Boolean`, `String`, or `Number` will be done when dealing with audience conditions.


tylerbrandt:
build

mikeng13:
build-e2e

core-api/src/main/java/com/optimizely/ab/bucketing/DecisionService.java
```diff
@@ -131,7 +131,7 @@ public DecisionService(@Nonnull Bucketer bucketer,
         if (ExperimentUtils.isUserInExperiment(projectConfig, experiment, filteredAttributes)) {
             String bucketingId = userId;
             if (filteredAttributes.containsKey(ControlAttribute.BUCKETING_ATTRIBUTE.toString())) {
-                bucketingId = filteredAttributes.get(ControlAttribute.BUCKETING_ATTRIBUTE.toString());
+                bucketingId = (String) filteredAttributes.get(ControlAttribute.BUCKETING_ATTRIBUTE.toString());
```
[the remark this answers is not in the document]
mikeng13: +1

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -45,12 +45,12 @@ public String getType() {
...
 
-    public String getValue() {
+    public Object getValue() {
         return value;
     }
 
-    public boolean evaluate(Map<String, String> attributes) {
-        String userAttributeValue = attributes.get(name);
+    public boolean evaluate(Map<String, ?> attributes) {
+        Object userAttributeValue = attributes.get(name);
```
mikeng13: Hmm should be careful with this. Eventually when we do accept objects and arrays to be passed in, the `equals` evaluation below will likely perform a reference check as well. That said, I think it's fine for now, but just leave a message saying that we should be extra considerate when we decide to allow objects and arrays.

core-api/src/main/java/com/optimizely/ab/bucketing/DecisionService.java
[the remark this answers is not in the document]
shihpatrick: If it isn't a string, should we keep it as `userId`?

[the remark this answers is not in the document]
mikeng13: Correct, we don't use the `bucketingId` if it's not a string and instead keep using the `userId`

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -765,7 +766,7 @@ public UserProfileService getUserProfileService() {
         if (unknownAttributes != null) {
             logger.warn("Attribute(s) {} not in the datafile.", unknownAttributes);
             // make a copy of the passed through attributes, then remove the unknown list
-            attributes = new HashMap<String, String>(attributes);
+            attributes = new HashMap<String, Object>(attributes);
```
wangjoshuah: You should add a test that actually removes unknown attributes to see what happens here. But my understanding is that `Object` is actually different from `?`. 
See https://stackoverflow.com/questions/678822/what-is-the-difference-between-and-object-in-java-generics

thomaszurkan-optimizely: You can't use a question mark in a constructor.

thomaszurkan-optimizely: I actually think that the proper way to use the constructor is HashMap<>(attributes).  So, @wangjoshuah is correct.

Please update to something like:
HashMap<String, ?> attr = new HashMap<>();

shihpatrick: Looks like `<>` was not introduced until Java 7 and we're currently on Java 6. Thoughts? 

https://stackoverflow.com/questions/16612157/why-does-new-hashmap-produce-an-error-in-jdk-1-6-but-not-1-7?lq=1

nchilada: Okay, it sounds like
```
new HashMap<String, Object>();
```
is indeed the way to go...

core-api/src/main/java/com/optimizely/ab/bucketing/DecisionService.java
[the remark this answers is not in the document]
nchilada: And I assume you're on top of this, but we should log a ~warning if we're ignoring the `bucketingId` due to type mismatch.

And while you're at it...we should also log at ~info level in cases where do ingest a usable `bucketingId`!

core-api/src/main/java/com/optimizely/ab/Optimizely.java
wangjoshuah: https://github.com/optimizely/java-sdk/pull/208 is the way to go. Good call @thomaszurkan-optimizely

shihpatrick:
build

shihpatrick:
build

thomaszurkan-optimizely:
build

mikeng13:
build

shihpatrick:
build

shihpatrick:
### Testing the changes
This e2e test does not have the proper configuration -- it is built here https://jenkins.dz.optimizely.com/job/fullstack-e2e-pipeline/4318/ . The e2e tests aren't passing but are currently being fixed. 

### Testing the original code
I pieced together some material to hopefully override the current failing of the e2e tests. 
I wanted to see if the errors are consistent with the original `master`s of both the `java-sdk` and `java-testapp`. 
 - ran a [e2e test](https://jenkins.dz.optimizely.com/job/fullstack-e2e-pipeline/4320/console) with parameters:
   - `master` for `java-sdk`
   - [`pshih/revert-func-signatures`](https://github.com/optimizely/java-testapp/tree/pshih/revert-func-signatures) for `java-testapp`

### Comparison
If we compare the consoles of the two tests, we would notice that the errors from the first test that includes changes to attributes are the same as the errors from the original code (but there is one more error that shows up in the original code*). This _may_ suggest that the changes in this PR does not cause the errors.

 *The extra error caused is with `tests/test_user_profiles.py:285`

@mikeng13 @thomaszurkan-optimizely

core-api/src/main/java/com/optimizely/ab/bucketing/DecisionService.java
```diff
@@ -131,7 +131,13 @@ public DecisionService(@Nonnull Bucketer bucketer,
         if (ExperimentUtils.isUserInExperiment(projectConfig, experiment, filteredAttributes)) {
             String bucketingId = userId;
             if (filteredAttributes.containsKey(ControlAttribute.BUCKETING_ATTRIBUTE.toString())) {
-                bucketingId = filteredAttributes.get(ControlAttribute.BUCKETING_ATTRIBUTE.toString());
+                if (String.class.isInstance(filteredAttributes.get(ControlAttribute.BUCKETING_ATTRIBUTE.toString()))) {
+                    bucketingId = (String) filteredAttributes.get(ControlAttribute.BUCKETING_ATTRIBUTE.toString());
+                    logger.info("bucketingId is valid");
+                } 
+                else {
+                    logger.warn("bucketingID has type mismatch, defaulted to userId ");
```
mikeng13: nit: This should just say `BucketingID attribute is not a string. Defaulted to userId.`

core-api/src/main/java/com/optimizely/ab/bucketing/DecisionService.java
```diff
@@ -131,7 +131,13 @@ public DecisionService(@Nonnull Bucketer bucketer,
         if (ExperimentUtils.isUserInExperiment(projectConfig, experiment, filteredAttributes)) {
             String bucketingId = userId;
             if (filteredAttributes.containsKey(ControlAttribute.BUCKETING_ATTRIBUTE.toString())) {
-                bucketingId = filteredAttributes.get(ControlAttribute.BUCKETING_ATTRIBUTE.toString());
+                if (String.class.isInstance(filteredAttributes.get(ControlAttribute.BUCKETING_ATTRIBUTE.toString()))) {
+                    bucketingId = (String) filteredAttributes.get(ControlAttribute.BUCKETING_ATTRIBUTE.toString());
+                    logger.info("bucketingId is valid");
```
mikeng13: nit: would be useful to print out what that bucketing ID is. Also, this should be a `debug` message.

core-api/src/main/java/com/optimizely/ab/bucketing/DecisionService.java
```diff
@@ -212,7 +218,13 @@ public DecisionService(@Nonnull Bucketer bucketer,
         int rolloutRulesLength = rollout.getExperiments().size();
         String bucketingId = userId;
         if (filteredAttributes.containsKey(ControlAttribute.BUCKETING_ATTRIBUTE.toString())) {
-            bucketingId = filteredAttributes.get(ControlAttribute.BUCKETING_ATTRIBUTE.toString());
+            if (String.class.isInstance(filteredAttributes.get(ControlAttribute.BUCKETING_ATTRIBUTE.toString()))) {
```
mikeng13: This logic is repeated from up above. Can you consolidate into a helper function?

core-api/src/main/java/com/optimizely/ab/config/audience/Condition.java
```diff
@@ -23,5 +23,5 @@
  */
 public interface Condition {
 
-    boolean evaluate(Map<String, String> attributes);
+    boolean evaluate(Map<String, ?> attributes);
```
mikeng13: Update the header on this file to be `2016-2018`. Please also do it for other files you've touched in this commit.

core-api/src/main/java/com/optimizely/ab/config/audience/NotCondition.java
```diff
@@ -37,7 +37,7 @@ public Condition getCondition() {
         return condition;
     }
 
-    public boolean evaluate(Map<String, String> attributes) {
+    public boolean evaluate(Map<String, ?> attributes) {
```
mikeng13: Do we need to update the `toString` method on this class to account for non-string values? Seems like it could lead to exceptions

core-api/src/main/java/com/optimizely/ab/config/audience/OrCondition.java
```diff
@@ -36,7 +36,7 @@ public OrCondition(@Nonnull List<Condition> conditions) {
         return conditions;
     }
 
-    public boolean evaluate(Map<String, String> attributes) {
+    public boolean evaluate(Map<String, ?> attributes) {
```
mikeng13: Same as above re: the toString method

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -29,9 +29,9 @@
 
     private final String name;
     private final String type;
-    private final String value;
+    private final Object value;
 
-    public UserAttribute(@Nonnull String name, @Nonnull String type, @Nullable String value) {
+    public UserAttribute(@Nonnull String name, @Nonnull String type, @Nullable Object value) {
```
mikeng13: @thomaszurkan-optimizely do we need to overload this constructor to allow you to pass in a `String value`? I don't think this change will break current usage of the class.

core-api/src/test/java/com/optimizely/ab/OptimizelyTest.java
```diff
@@ -43,6 +43,7 @@
 import com.optimizely.ab.notification.NotificationCenter;
 import com.optimizely.ab.notification.TrackNotificationListener;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import org.hamcrest.Matcher;
```
mikeng13: Is this still needed? Maybe I am missing where it's used

core-api/src/test/java/com/optimizely/ab/notification/NotificationCenterTest.java
```diff
@@ -70,7 +70,7 @@ public void onActivate(@Nonnull Experiment experiment, @Nonnull String userId, @
     public void testAddActivateNotification() {
         int notificationId = notificationCenter.addActivateNotificationListener(new ActivateNotificationListenerInterface() {
             @Override
-            public void onActivate(@Nonnull Experiment experiment, @Nonnull String userId, @Nonnull Map<String, String> attributes, @Nonnull Variation variation, @Nonnull LogEvent event) {
+            public void onActivate(@Nonnull Experiment experiment, @Nonnull String userId, @Nonnull Map<String, ?> attributes, @Nonnull Variation variation, @Nonnull LogEvent event) {
```
mikeng13: Will this break current usage of the notification listeners? If someone has already implemented notification listeners with just `Map<String, String>`, will they suddenly see build errors?

shihpatrick: It would break current usage (I had to make changes in unit test and the java-testapp). Does it make sense to leave it as Map<String, String> though?

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
mikeng13: Nevermind, I think this will be fine as is. Can ignore my previous comment.

core-api/src/test/java/com/optimizely/ab/notification/NotificationCenterTest.java
mikeng13: Let's hold off on this until we discuss it today. My inclination is that we leave `Map<String, String>`, and in our notification center, coerce all values to String types

core-api/src/main/java/com/optimizely/ab/bucketing/DecisionService.java
```diff
@@ -343,4 +337,25 @@ void saveVariation(@Nonnull Experiment experiment,
             }
         }
     }
+
+    /**
+     * @param userId The userId of the user.
```
mikeng13: Please provide a description for this method. Also, please make it private as it is not used outside this class.

core-api/src/main/java/com/optimizely/ab/bucketing/DecisionService.java
```diff
@@ -129,10 +129,7 @@ public DecisionService(@Nonnull Bucketer bucketer,
         }
 
         if (ExperimentUtils.isUserInExperiment(projectConfig, experiment, filteredAttributes)) {
-            String bucketingId = userId;
-            if (filteredAttributes.containsKey(ControlAttribute.BUCKETING_ATTRIBUTE.toString())) {
-                bucketingId = filteredAttributes.get(ControlAttribute.BUCKETING_ATTRIBUTE.toString());
-            }
+            String bucketingId = getBucketingId(userId,filteredAttributes);
```
mikeng13: nit: space after `filteredAttributes`

core-api/src/main/java/com/optimizely/ab/config/audience/Audience.java
```diff
@@ -1,6 +1,6 @@
 /**
  *
- *    Copyright 2016-2017, Optimizely and contributors
+ *    Copyright 2016-2018, Optimizely and contributors
```
mikeng13: You didn't change this class so no need to update the year.

core-api/src/test/java/com/optimizely/ab/notification/NotificationCenterTest.java
shihpatrick: From today's discussion -- looks like we're using `Map<String, ?>`

mikeng13: Yep we are.

core-api/src/main/java/com/optimizely/ab/notification/NotificationListener.java
```diff
@@ -1,6 +1,6 @@
 /**
  *
- *    Copyright 2016-2017, Optimizely and contributors
+ *    Copyright 2016-2018, Optimizely and contributors
```
mikeng13: Please also revert this change since we aren't touching this file this year

PR #209, opened 2018-08-31
refactor(init): build invalid instance if datafile cannot be parsed.

mikeng13:
## Summary

Refactor the Java SDK to follow the rest of the Full Stack SDKs in not throwing or preventing SDK initialization when an invalid datafile is passed in. Instead we will return an `Optimizely` instance with the `isValid` property set to `false`.  When accessing any of the SDK APIs, we will first check if the instance is valid before proceeding. If not valid, we will return the sensible defaults for each API.

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/Optimizely.java
+++ b/core-api/src/main/java/com/optimizely/ab/Optimizely.java
@@ -29,7 +29,6 @@ import com.optimizely.ab.config.LiveVariableUsageInstance;
 import com.optimizely.ab.config.ProjectConfig;
 import com.optimizely.ab.config.Variation;
 import com.optimizely.ab.config.parser.ConfigParseException;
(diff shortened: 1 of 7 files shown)


core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -80,33 +79,55 @@
...
 
-    // Do work here that should be done once per Optimizely lifecycle
+    /**
+     * Initializes the SDK state. Can conceivably re-use this in the future with datafile sync where
+     * we can re-initialize the SDK instead of re-instantiating.
+     */
     @VisibleForTesting
-    void initialize() {
+    void initialize(@Nonnull String datafile, @Nullable ProjectConfig projectConfig) {
+        if (projectConfig == null) {
```
mikeng13: This step de-couples the "building" of the Optimizely instance from the "initialization", which is where we start populating state.

PR #212, opened 2018-09-10
fix(nullfilter): Filters attributes in getFeatureVariableValueForType 

mnoman09:
Filters attributes getFeatureVariableValueForType before passing it to decisionService.getVariationForFeature So if attributes passed is null than it will handle it.


mikeng13:
Closing this as well as it is taken care of in #214

PR #213, opened 2018-09-10
New audience match types

thomaszurkan-optimizely:
- New match types along with leaf matchers
- Changed logic of AND,OR,NOT conditions to allow for nulls to bubble through appropriately.  
- Change audience evaluation to always evaluate regardless if there was an attribute key passed in or not.
- Still allows null in your attribute list.
- Added match to UserAttribute in preparation of typedAttributes.  Added to all json parsers as a default.
- Parse typedAudiences and include in v4 datafile validation. V3 and V2


core-api/src/main/java/com/optimizely/ab/config/audience/MatchType.java
```diff
@@ -0,0 +1,187 @@
...
+        return value.equals(convert(otherValue));
+    }
+}
+
+/**
+ * This is a temporary class.  It mimics the current behaviour for
+ * custom dimension.  This will be dropped for ExactMatch and the unit tests need to be fixed.
+ * @param <T>
+ */
+class CustomDimensionMatch<T> extends LeafMatch<T> {
```
mikeng13: We don't support custom dimensions anymore, even in 1.x. I think we can safely omit this.

core-api/src/main/java/com/optimizely/ab/config/audience/MatchType.java
```diff
@@ -0,0 +1,187 @@
...
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+interface LeafMatcher {
+    public Boolean eval(Object otherValue);
+}
+
+abstract class LeafMatch<T> implements LeafMatcher {
+    T convert(Object o) {
```
mikeng13: Please document this method.

mikeng13: Also, I think we should be letting these throw so we can log the type mismatch as a debug level log?

core-api/src/main/java/com/optimizely/ab/config/audience/MatchType.java
```diff
@@ -0,0 +1,187 @@
...
+    }
+}
+
+public class MatchType {
+
+    private String type;
+    private LeafMatcher matcher;
+
+    public static MatchType getMatchType(String type, Object value) {
+        if (type == null) type = "custom_dimension";
```
mikeng13: I am not sure this is accurate, we should default to `exact` right?

core-api/src/main/java/com/optimizely/ab/config/audience/MatchType.java
```diff
@@ -0,0 +1,187 @@
...
+
+class NullMatch extends LeafMatch<Object> {
+    @SuppressFBWarnings("URF_UNREAD_FIELD")
+    Object value;
+    protected NullMatch() {
+        this.value = null;
+    }
+
+    public @Nullable Boolean eval(Object otherValue) {
+        return null;
```
mikeng13: This is a bit confusing, shouldn't this match the value to `null`?

core-api/src/main/java/com/optimizely/ab/config/audience/OrCondition.java
```diff
@@ -36,10 +37,22 @@ public OrCondition(@Nonnull List<Condition> conditions) {
         return conditions;
     }
 
-    public boolean evaluate(Map<String, ?> attributes) {
+    // https://docs.google.com/document/d/158_83difXVXF0nb91rxzrfHZwnhsybH21ImRA_si7sg/edit#
+    // According to the matix mentioned in the above document.
```
mikeng13: nit: typo in `matrix`

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -45,25 +47,29 @@ public String getType() {
...
+    public String getMatch() {
+        return match;
+    }
+
     public Object getValue() {
         return value;
     }
 
-    public boolean evaluate(Map<String, ?> attributes) {
+    public Boolean evaluate(Map<String, ?> attributes) {
```
mikeng13: Please add unit tests for this method.

core-api/src/main/java/com/optimizely/ab/config/audience/MatchType.java
```diff
@@ -0,0 +1,187 @@
+package com.optimizely.ab.config.audience;
```
mikeng13: Please add unit tests for the Match types as well as the different Conditions classes, especially around the `evaluate` logic.

core-api/src/main/java/com/optimizely/ab/config/parser/JsonConfigParser.java
```diff
@@ -289,13 +295,14 @@ private Condition parseConditions(JSONArray conditionJson) {
...
-                String value = null;
+                Object value = null;
                 if (conditionMap.has("value")) {
-                    value = conditionMap.getString("value");
+                    value = conditionMap.get("value");
                 }
                 conditions.add(new UserAttribute(
                         (String)conditionMap.get("name"),
                         (String)conditionMap.get("type"),
+                        conditionMap.has("match")?(String)conditionMap.get("match"):null,
```
mikeng13: nit: can we space out the ternary match for readability

core-api/src/main/java/com/optimizely/ab/config/parser/JsonSimpleConfigParser.java
```diff
@@ -71,6 +71,10 @@ public ProjectConfig parseProjectConfig(@Nonnull String json) throws ConfigParse
 
             List<EventType> events = parseEvents((JSONArray)rootObject.get("events"));
             List<Audience> audiences = parseAudiences((JSONArray)parser.parse(rootObject.get("audiences").toString()));
+            List<Audience> typedAudiences =null;
```
mikeng13: nit: space after `=`

core-api/src/main/java/com/optimizely/ab/internal/ExperimentUtils.java
```diff
@@ -66,14 +66,11 @@ public static boolean isUserInExperiment(@Nonnull ProjectConfig projectConfig,
             return true;
         }
 
-        // if there are audiences, but no user attributes, the user is not in the experiment.
```
mikeng13: Nice :)

core-api/src/main/java/com/optimizely/ab/internal/ExperimentUtils.java
```diff
@@ -66,14 +66,11 @@ public static boolean isUserInExperiment(@Nonnull ProjectConfig projectConfig,
...
-        if (attributes.isEmpty()) {
-            return false;
-        }
-
         for (String audienceId : experimentAudienceIds) {
             Condition conditions = projectConfig.getAudienceConditionsFromId(audienceId);
-            if (conditions.evaluate(attributes)) {
+            Boolean conditionEval = conditions.evaluate(attributes);
+
+            if (conditionEval != null && conditionEval){
```
mikeng13: Can this be shortened to if `conditionEval == true`?

core-api/src/test/java/com/optimizely/ab/config/ProjectConfigTest.java
```diff
@@ -106,7 +106,7 @@ public void verifyGetExperimentsForInvalidEvent() throws Exception {
     @Test
     public void verifyGetAudienceConditionsFromValidId() throws Exception {
         List<Condition> userAttributes = new ArrayList<Condition>();
-        userAttributes.add(new UserAttribute("browser_type", "custom_attribute", "firefox"));
+        userAttributes.add(new UserAttribute("browser_type", "custom_attribute", null,"firefox"));
```
mikeng13: nit: space after `null`

core-api/src/test/java/com/optimizely/ab/config/ProjectConfigTestUtils.java
```diff
@@ -245,7 +245,7 @@ private static ProjectConfig generateValidProjectConfigV3() {
                                         new EventType("100", "no_running_experiments", singletonList("118")));
 
         List<Condition> userAttributes = new ArrayList<Condition>();
-        userAttributes.add(new UserAttribute("browser_type", "custom_attribute", "firefox"));
+        userAttributes.add(new UserAttribute("browser_type", "custom_attribute", null,"firefox"));
```
mikeng13: nit: space after `null`

core-api/src/main/java/com/optimizely/ab/internal/ExperimentUtils.java
thomaszurkan-optimizely: Need to check for null.  Auto boxing I believe would throw an exception here.

core-api/src/main/java/com/optimizely/ab/config/audience/match/CustomDimensionMatch.java
```diff
@@ -0,0 +1,36 @@
...
+package com.optimizely.ab.config.audience.match;
+
+import javax.annotation.Nullable;
+
+/**
+ * This is a temporary class.  It mimics the current behaviour for
+ * custom dimension.  This will be dropped for ExactMatch and the unit tests need to be fixed.
+ * @param <T>
+ */
+class CustomDimensionMatch<T> extends LeafMatch<T> {
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: If there is no match we use the old style matcher which is captured in CustomDimension (this is before we changed everything to custom attribute.  Do you think it should be named DefaultMatchForLegacyAttributes?

thomaszurkan-optimizely:
build

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -45,25 +49,29 @@ public String getType() {
...
-    public boolean evaluate(Map<String, ?> attributes) {
+    public @Nullable Boolean evaluate(Map<String, ?> attributes) {
         // Valid for primative types, but needs to change when a value is an object or an array
         Object userAttributeValue = attributes.get(name);
 
-        if (value != null) { // if there is a value in the condition
-            // check user attribute value is equal
-            return value.equals(userAttributeValue);
+        if (!"custom_attribute".equals(type)) {
+            return null; // unknown type
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: added logging instead.

core-api/src/main/java/com/optimizely/ab/config/audience/OrCondition.java
```diff
@@ -36,10 +37,25 @@ public OrCondition(@Nonnull List<Condition> conditions) {
...
+            Boolean conditionEval = condition.evaluate(attributes);
+            if (conditionEval == null) { // true with falses and nulls is still true
+                foundNull = true;
+            }
+            else if (conditionEval) {
                 return true;
+            }
+        }
+
+        if (foundNull) {// if found null and false return null.  all false return false
```
mikeng13: nit: space before the start of the comment

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -45,25 +49,31 @@ public String getType() {
...
 
-    public boolean evaluate(Map<String, ?> attributes) {
+    public @Nullable Boolean evaluate(Map<String, ?> attributes) {
         // Valid for primative types, but needs to change when a value is an object or an array
         Object userAttributeValue = attributes.get(name);
 
-        if (value != null) { // if there is a value in the condition
-            // check user attribute value is equal
-            return value.equals(userAttributeValue);
+        if (!"custom_attribute".equals(type)) {
```
mikeng13: nit: Why not invoke the `equals` method from the variable itself?

core-api/src/main/java/com/optimizely/ab/config/audience/match/GTMatch.java
```diff
@@ -0,0 +1,37 @@
...
+        this.value = value;
+    }
+
+    public @Nullable
+    Boolean eval(Object otherValue) {
+        try {
+            return convert(otherValue).doubleValue() > value.doubleValue();
+        }
+        catch (Exception e) {
+            MatchType.logger.error("Greater than match ", e);
```
mikeng13: nit: Can be a bit more verbose: `Greater than match failed`

core-api/src/main/java/com/optimizely/ab/config/audience/match/LTMatch.java
```diff
@@ -0,0 +1,38 @@
...
+        this.value = value;
+    }
+
+    public @Nullable
+    Boolean eval(Object otherValue) {
+        try {
+            return convert(otherValue).doubleValue() < value.doubleValue();
+        }
+        catch (Exception e) {
+            MatchType.logger.error("Less than match ", e);
```
mikeng13: Same as above

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchType.java
```diff
@@ -0,0 +1,90 @@
...
+                if (value instanceof Number) {
+                    return new MatchType(type, new GTMatch((Number) value));
+                }
+                break;
+            case "lt":
+                if (value instanceof Number) {
+                    return new MatchType(type, new LTMatch((Number) value));
+                }
+                break;
+            case "legacy_custom_attribute":
```
mikeng13: Huh? Is this something that we are populating in the datafile? @nchilada I don't recall seeing this in the design doc. Isn't this the default case?

core-api/src/main/java/com/optimizely/ab/config/audience/match/SubstringMatch.java
```diff
@@ -0,0 +1,38 @@
...
+        this.value = value;
+    }
+
+    public @Nullable
+    Boolean eval(Object otherValue) {
+        try {
+            return value.contains(convert(otherValue));
+        }
+        catch (Exception e) {
+            MatchType.logger.error("Substring match ", e);
```
mikeng13: `Substring match failed`

core-api/src/main/java/com/optimizely/ab/internal/ExperimentUtils.java
mikeng13: OK. 

Also, nit: space before the opening `{`

core-api/src/main/java/com/optimizely/ab/config/audience/match/GTMatch.java
thomaszurkan-optimizely: the exception will be included in the output making it pretty obvious it is an error.

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchType.java
thomaszurkan-optimizely: No, I changed the legacy so that if no match type is provide, I try and use the legacy custom attribute matcher.

thomaszurkan-optimizely: It's set at the top of the method when match is null (name="browser", type="custom_attribute" value="firefox") would end up using legacy.  I thought changing the name would make that clearer. :(

.travis.yml
```diff
@@ -26,3 +26,6 @@ branches:
   - /^\d+\.\d+\.\d+(-SNAPSHOT|-alpha|-beta)?\d*$/ # trigger builds on tags which are semantically versioned to ship the SDK.
 after_success:
   - ./gradlew coveralls uploadArchives --console plain
+after_failure:
+  - cat /home/travis/build/optimizely/java-sdk/core-api/build/reports/findbugs/main.html
+
```
aliabbasrizvi: nit. Remove extraneous line.

core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
```diff
@@ -170,6 +173,14 @@ public ProjectConfig(String accountId,
 
         this.attributes = Collections.unmodifiableList(attributes);
         this.audiences = Collections.unmodifiableList(audiences);
+
+        if (typedAudiences != null) {
+            this.typedAudiences = Collections.unmodifiableList(typedAudiences);
+        }
+        else {
+            this.typedAudiences = Collections.emptyList();
```
aliabbasrizvi: Wouldn't it make life easier if `this.typedAudiences = Collections.unmodifiableList(audiences)` instead?

core-api/src/main/java/com/optimizely/ab/config/audience/Condition.java
```diff
@@ -16,12 +16,14 @@
...
 
 /**
  * Interface implemented by all conditions condition objects to aid in condition evaluation.
  */
 public interface Condition {
 
-    boolean evaluate(Map<String, ?> attributes);
+    @Nullable
+    Boolean evaluate(Map<String, ?> attributes);
 }
```
aliabbasrizvi: nit. Insert new line here.

core-api/src/test/resources/config/valid-project-config-v4.json
```diff
@@ -114,6 +151,36 @@
...
+        {
+          "entityId": "1423767503",
+          "endOfRange": 5000
+        },
+        {
+          "entityId": "3433458315",
+          "endOfRange": 10000
+        }
+      ],
+      "audienceIds": ["3468206643","3468206644","3468206645"],
```
aliabbasrizvi: nit. Insert space after ,

core-api/src/main/java/com/optimizely/ab/config/audience/match/DefaultMatchForLegacyAttributes.java
```diff
@@ -0,0 +1,36 @@
...
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+package com.optimizely.ab.config.audience.match;
+
+import javax.annotation.Nullable;
+
+/**
+ * This is a temporary class.  It mimics the current behaviour for
+ * legasy custom attributes.  This will be dropped for ExactMatch and the unit tests need to be fixed.
```
aliabbasrizvi: typo. legacy

core-api/src/main/java/com/optimizely/ab/config/audience/match/LeafMatch.java
```diff
@@ -0,0 +1,28 @@
...
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+package com.optimizely.ab.config.audience.match;
+
+abstract class LeafMatch<T> implements LeafMatcher {
```
aliabbasrizvi: Not a big fan of this name `LeafMatch`. May be just `Match`?

core-api/src/main/java/com/optimizely/ab/config/audience/match/LeafMatcher.java
```diff
@@ -0,0 +1,21 @@
...
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+package com.optimizely.ab.config.audience.match;
+
+public interface LeafMatcher {
```
aliabbasrizvi: Same feedback as above. Perhaps just call this `Matcher`

.travis.yml
thomaszurkan-optimizely: this has been too useful to remove.  when travis fails findbugs for a JDK other than the one you are working on, this becomes invaluable in finding the reason for the failure.

core-api/src/main/java/com/optimizely/ab/config/parser/AudienceJacksonDeserializer.java
```diff
@@ -59,19 +59,23 @@ private Condition parseConditions(List<Object> rawObjectList) {
...
+        switch (operand) {
+            case "and":
+                condition = new AndCondition(conditions);
+                break;
+            case "or":
+                condition = new OrCondition(conditions);
+                break;
+            default:
+                condition = new NotCondition(conditions.get(0));
+                break;
```
nchilada: Can we add support for implicit "or" operators as outlined in the spec?

core-api/src/main/java/com/optimizely/ab/internal/ExperimentUtils.java
nchilada: :yasss:

core-api/src/main/java/com/optimizely/ab/internal/ExperimentUtils.java
```diff
@@ -66,14 +66,11 @@ public static boolean isUserInExperiment(@Nonnull ProjectConfig projectConfig,
...
-        if (attributes.isEmpty()) {
-            return false;
-        }
-
         for (String audienceId : experimentAudienceIds) {
             Condition conditions = projectConfig.getAudienceConditionsFromId(audienceId);
-            if (conditions.evaluate(attributes)) {
+            Boolean conditionEval = conditions.evaluate(attributes);
+
+            if (conditionEval != null && conditionEval) {
```
nchilada: This looks good for now! But I assume this is the part we'll be working on next

core-api/src/test/resources/config/valid-project-config-v4.json
```diff
@@ -27,6 +27,43 @@
...
+      "id": "4194404272",
+      "name": "english_citizens",
+      "conditions": "[\"and\", [\"or\", [\"or\", {\"name\": \"nationality\", \"type\": \"custom_attribute\", \"match\":\"exact\", \"value\":\"English\"}]]]"
+    },
+    {
+      "id": "2196265320",
+      "name": "audience_with_missing_value",
+      "conditions": "[\"and\", [\"or\", [\"or\", {\"name\": \"nationality\", \"type\": \"custom_attribute\", \"value\": \"English\"}, {\"name\": \"nationality\", \"type\": \"custom_attribute\"}]]]"
+    }
+  ],
```
nchilada: Since you'd asked me to check: this looks accurate. But I'd also test with an audience whose ID is found in both `audiences` and `typedAudiences`.

It might also be wise to include audience conditions with `match` being `null` instead of being undefined, though I have no _intention_ of having the app backend serialize such conditions. 🤷

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
nchilada: Good thinking @thomaszurkan-optimizely!

nchilada: Once we support other condition types we'll probably want to do this check in ProjectConfig or wherever. But this is fine for now. I'm not overly concerned about the fact that we're constructing CustomAttribute objects for things that might not semantically represent `custom_attribute` conditions.

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -45,25 +49,31 @@ public String getType() {
...
+            MatchType.logger.error(String.format("condition type not equal to `custom_attribute` %s", type != null ? type : ""));
+            return null; // unknown type
         }
-        else if (userAttributeValue != null) { // if the datafile value is null but user has a value for this attribute
-            // return false since null != nonnull
-            return false;
+        // check user attribute value is equal
+        try {
+            return MatchType.getMatchType(match, value).getMatcher().eval(userAttributeValue);
```
nchilada: Based on the name alone, I find it surprising that we're passing `value` to `.getMatchType(...)` and not to `.getMatcher(...)` or `.eval(...)`. That said, this is minor and I haven't looked at those functions yet.

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -45,25 +49,31 @@ public String getType() {
...
         }
-        else { // both are null
-            return true;
+        catch (NullPointerException np) {
+            MatchType.logger.error(String.format("attribute or value null for match %s", match != null ? match : "legacy condition"),np);
```
nchilada: We can definitely wait until after the initial PR is merged, but we may want to have super-specific error messages. cc @ceimaj

core-api/src/main/java/com/optimizely/ab/config/audience/match/AttributeMatch.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+package com.optimizely.ab.config.audience.match;
+
+abstract class AttributeMatch<T> implements Match {
+    T convert(Object o) {
+        try {
+            T rv = (T)o;
+            return rv;
+        } catch(java.lang.ClassCastException e) {
+            return null;
```
nchilada: Swallowing the error and returning `null` feels pretty weird to me. This causes subclasses' `eval(...)` methods to proceed blindly with a `null` value (in some cases producing `NullPointerException`s which they must know to catch) when the more straightforward, robust behavior might be for them to immediately return `null`.

Basically, I think the pattern smells.

As one alternative: what if we tried something like
```java
abstract class TypedMatch<T> implements Match {
    Boolean eval(Object otherValue) {
        try {
            T otherValueT = (T) otherValue;
        } catch(java.lang.ClassCastException e) {
            return null;
        }
        return eval(typedOtherValue);
    }

    Boolean eval(T otherValue);
}
```
and had our subclasses the new, typed abstract method instead?

(Or give up on having common, generic-powered type-checking code? But maybe that's just my dynamic typing background coming through 🤷)

core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
```diff
@@ -206,7 +217,14 @@ public ProjectConfig(String accountId,
...
 
         // generate audience id to audience mapping
-        this.audienceIdMapping = ProjectConfigUtils.generateIdMapping(audiences);
+        if (typedAudiences == null) {
+            this.audienceIdMapping = ProjectConfigUtils.generateIdMapping(audiences);
+        }
+        else {
+            List<Audience> combinedList = new ArrayList<>(audiences);
+            combinedList.addAll(typedAudiences);
+            this.audienceIdMapping = ProjectConfigUtils.generateIdMapping(combinedList);
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: yes. that is exactly the point of this combination.  the map is audience ids.  The first set of mappings is for audiences.  If the Id exists, it is replaced when the typed audiences is added to the map.

core-api/src/main/java/com/optimizely/ab/config/parser/AudienceGsonDeserializer.java
```diff
@@ -67,19 +67,23 @@ private Condition parseConditions(List<Object> rawObjectList) {
...
+        switch (operand) {
+            case "and":
+                condition = new AndCondition(conditions);
+                break;
+            case "or":
+                condition = new OrCondition(conditions);
+                break;
+            default:
+                condition = new NotCondition(conditions.get(0));
+                break;
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: Not in this PR

core-api/src/main/java/com/optimizely/ab/config/parser/AudienceJacksonDeserializer.java
thomaszurkan-optimizely: Not in this PR

core-api/src/main/java/com/optimizely/ab/config/parser/JsonConfigParser.java
```diff
@@ -69,6 +69,11 @@ public ProjectConfig parseProjectConfig(@Nonnull String json) throws ConfigParse
 
             List<EventType> events = parseEvents(rootObject.getJSONArray("events"));
             List<Audience> audiences = parseAudiences(rootObject.getJSONArray("audiences"));
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: Why?

core-api/src/main/java/com/optimizely/ab/config/parser/JsonConfigParser.java
```diff
@@ -289,25 +295,27 @@ private Condition parseConditions(JSONArray conditionJson) {
...
+        switch (operand) {
+            case "and":
+                condition = new AndCondition(conditions);
+                break;
+            case "or":
+                condition = new OrCondition(conditions);
+                break;
+            default:
+                condition = new NotCondition(conditions.get(0));
+                break;
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: Not in this PR

core-api/src/main/java/com/optimizely/ab/config/parser/JsonSimpleConfigParser.java
```diff
@@ -298,7 +305,7 @@ private Condition parseConditions(JSONArray conditionJson) {
             } else {
                 JSONObject conditionMap = (JSONObject)obj;
                 conditions.add(new UserAttribute((String)conditionMap.get("name"), (String)conditionMap.get("type"),
-                               (String)conditionMap.get("value")));
+                        (String)conditionMap.get("match"), conditionMap.get("value")));
             }
         }
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: Not in this PR.  This is just typed audiences.

core-api/src/test/resources/config/valid-project-config-v4.json
thomaszurkan-optimizely: All the string conditions that map to exact are also in the audience list and replace the audience lookup map with the typed audience since they share ids.  Look at the audience list above typed audience definition.

core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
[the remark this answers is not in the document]
nchilada: Okay thanks. That's quite implicit but I guess it works. A code comment and unit tests would help!

core-api/src/main/java/com/optimizely/ab/config/parser/AudienceGsonDeserializer.java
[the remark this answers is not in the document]
nchilada: Ah, because we weren't doing that before? I see that now.

Happy to defer but in that case I'll file a ticket to track that change separately for the same release

core-api/src/main/java/com/optimizely/ab/config/parser/JsonConfigParser.java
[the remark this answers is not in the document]
nchilada: So that app backend isn't tied down with `audiences` if we decide to move entirely to `typedAudiences` at some point. Semantically, it seems reasonable to make `audiences` optional once `typedAudiences` is supported too.

core-api/src/test/resources/config/valid-project-config-v4.json
nchilada: Ah, gotcha. I should've expanded the whole file and checked 😬

core-api/src/main/java/com/optimizely/ab/config/parser/AudienceGsonDeserializer.java
```diff
@@ -67,19 +67,23 @@ private Condition parseConditions(List<Object> rawObjectList) {
...
-            condition = new NotCondition(conditions.get(0));
+        switch (operand) {
+            case "and":
+                condition = new AndCondition(conditions);
+                break;
+            case "or":
+                condition = new OrCondition(conditions);
+                break;
+            default:
+                condition = new NotCondition(conditions.get(0));
```
nchilada: After discussing with @mjc1283: we should also return `null` instead of raising `IndexOutOfBoundsException` if `conditions.length == 0`.

core-api/src/main/java/com/optimizely/ab/config/audience/match/AttributeMatch.java
```diff
@@ -0,0 +1,28 @@
...
+ *    limitations under the License.
+ */
+package com.optimizely.ab.config.audience.match;
+
+abstract class AttributeMatch<T> implements Match {
+    T convert(Object o) {
+        try {
+            T rv = (T)o;
+            return rv;
+        } catch(java.lang.ClassCastException e) {
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: Adding logging

core-api/src/main/java/com/optimizely/ab/config/audience/match/AttributeMatch.java
thomaszurkan-optimizely: This is a great idea and I really like where you are going with it.  However, in Java (and other strongly typed languages) the eval here ends up with the same signature.  Meaning a value is both of type object and type t.  So, it becomes ambiguous.  :(

core-api/src/main/java/com/optimizely/ab/config/parser/JsonConfigParser.java
```diff
@@ -289,25 +295,27 @@ private Condition parseConditions(JSONArray conditionJson) {
                 conditions.add(parseConditions(conditionJson.getJSONArray(i)));
             } else {
                 JSONObject conditionMap = (JSONObject)obj;
-                String value = null;
-                if (conditionMap.has("value")) {
-                    value = conditionMap.getString("value");
-                }
+                Object value = conditionMap.has("value") ? conditionMap.get("value") : null;
+                String match = conditionMap.has("match") ? (String) conditionMap.get("match") : null;
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: The map type was changed from <String,String> to <String,?> so that the implicit getString() on an object value fails.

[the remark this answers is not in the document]
nchilada: Okay, thanks. I thought I'd seen continued usage of `getString` in similar context, but I'm not seeing any other usages in the current diff, and must have just misunderstood the difference. Thanks for explaining!

core-api/src/main/java/com/optimizely/ab/config/audience/match/AttributeMatch.java
nchilada: Ah, interesting, thanks.

core-api/src/main/java/com/optimizely/ab/config/parser/AudienceGsonDeserializer.java
thomaszurkan-optimizely: Is there ever going to be a audience without a condition (not to be confused with audience id list)?

core-api/src/main/java/com/optimizely/ab/config/parser/JsonConfigParser.java
```diff
@@ -68,7 +69,17 @@ public ProjectConfig parseProjectConfig(@Nonnull String json) throws ConfigParse
             attributes = parseAttributes(rootObject.getJSONArray("attributes"));
 
             List<EventType> events = parseEvents(rootObject.getJSONArray("events"));
-            List<Audience> audiences = parseAudiences(rootObject.getJSONArray("audiences"));
+            List<Audience> audiences = Collections.emptyList();
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: backward compatibility.  A v2 or V3 would possibly have problems with a null audience.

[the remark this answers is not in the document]
thomaszurkan-optimizely: Whereas typedAudience is new and will be null in V2 and V3

core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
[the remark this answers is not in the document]
thomaszurkan-optimizely: This is in a bunch of unit tests that I added with V4 datafile with typedAudiences.  They share Ids for the string exact ones (i.e. legacy and typed).  The audience id map is used to get the audience and uses the combined map in question.

core-api/src/main/java/com/optimizely/ab/config/parser/AudienceGsonDeserializer.java
nchilada: > Is there ever going to be a audience without a condition (not to be confused with audience id list)?

As we started to discuss in Slack: if an audience object has null or missing `conditions`, I think that's rather problematic and it's appropriate for the SDK to reject the datafile. On the other hand, if the `conditions` are empty (`[]`, `{}`, etc.) then it's fine to apply our recursive evaluation algorithm to that value.

PR #214, opened 2018-09-11
fix(nullfilter): filters out attributes with null values

mnoman09:
This will ignore attributes with null values on creation of impression or conversion event.

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/event/internal/EventFactory.java
+++ b/core-api/src/main/java/com/optimizely/ab/event/internal/EventFactory.java
@@ -171,7 +171,7 @@ public class EventFactory {
             // Filter down to the types of values we're allowed to track.
             // Don't allow Longs, BigIntegers, or BigDecimals - they /can/ theoretically be serialized as JSON numbers
             // but may take on values that can't be faithfully parsed by the backend.
-            // https://github.com/optimizely/avro-schemas/blob/baeba19a7bd227aae67026474aec239110c5801d/event_batch/src/main/avro/EventBatch.avsc#L22
+            // https://developers.optimizely.com/x/events/api/#Attribute
             if (entry.getValue() == null ||
                     !((entry.getValue() instanceof String) ||
                     (entry.getValue() instanceof Integer) ||
--- a/core-api/src/main/java/com/optimizely/ab/event/internal/EventFactory.java
+++ b/core-api/src/main/java/com/optimizely/ab/event/internal/EventFactory.java
@@ -168,12 +168,13 @@ public class EventFactory {
         List<Attribute> attributesList = new ArrayList<Attribute>();
 
         for (Map.Entry<String, ?> entry : attributes.entrySet()) {
-            //Ignore null value attribute
+            // Filters down to the types of values that are allowed.
+            // Don't allow Longs, BigIntegers, or BigDecimals - they /can/ theoretically be serialized as JSON numbers
             if (entry.getValue() == null ||
-                    (!(entry.getValue() instanceof String) &&
-                    !(entry.getValue() instanceof Integer) &&
-                    !(entry.getValue() instanceof Double) &&
-                    !(entry.getValue() instanceof Boolean))) {
+                    !((entry.getValue() instanceof String) ||
+                    (entry.getValue() instanceof Integer) ||
+                    (entry.getValue() instanceof Double) ||
+                    (entry.getValue() instanceof Boolean))) {
                 continue;
             }
 
--- a/core-api/src/main/java/com/optimizely/ab/event/internal/EventFactory.java
+++ b/core-api/src/main/java/com/optimizely/ab/event/internal/EventFactory.java
@@ -169,7 +169,11 @@ public class EventFactory {
 
         for (Map.Entry<String, ?> entry : attributes.entrySet()) {
             //Ignore null value attribute
-            if (entry.getValue() == null) {
+            if (entry.getValue() == null ||
+                    (!(entry.getValue() instanceof String) &&
+                    !(entry.getValue() instanceof Integer) &&
+                    !(entry.getValue() instanceof Double) &&
+                    !(entry.getValue() instanceof Boolean))) {
                 continue;
             }
 
--- a/core-api/src/main/java/com/optimizely/ab/event/internal/EventFactory.java
+++ b/core-api/src/main/java/com/optimizely/ab/event/internal/EventFactory.java
@@ -168,8 +168,10 @@ public class EventFactory {
         List<Attribute> attributesList = new ArrayList<Attribute>();
 
         for (Map.Entry<String, ?> entry : attributes.entrySet()) {
-            // Filters down to the types of values that are allowed.
+            // Filter down to the types of values we're allowed to track.
             // Don't allow Longs, BigIntegers, or BigDecimals - they /can/ theoretically be serialized as JSON numbers
+            // but may take on values that can't be faithfully parsed by the backend.
+            // https://github.com/optimizely/avro-schemas/blob/baeba19a7bd227aae67026474aec239110c5801d/event_batch/src/main/avro/EventBatch.avsc#L22
             if (entry.getValue() == null ||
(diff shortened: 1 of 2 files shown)


core-api/src/main/java/com/optimizely/ab/event/internal/EventFactory.java
```diff
@@ -168,6 +168,11 @@ public LogEvent createConversionEvent(@Nonnull ProjectConfig projectConfig,
         List<Attribute> attributesList = new ArrayList<Attribute>();
 
         for (Map.Entry<String, ?> entry : attributes.entrySet()) {
+            //Ignore null value attribute
+            if (entry.getValue() == null) {
```
nchilada: Can we whitelist strings, numbers (probably just `int`s and `double`s for now, but the team might need to discuss more), and booleans rather than blacklisting some of the values we don't want to send? I should mention that OASIS-3537 has been updated and might not have mentioned that initially. 😬

msohailhussain: @nchilada this is related to OASIS-3535 not for OASIS-3537

nchilada: Oh, interesting. @mikeng13 can you clarify the difference between 3535 and 3537?

msohailhussain: As per my understanding, this is only related to NULL values, while OASIS-3537 is more related to filtering out all unsupported values but we have hold-off for OASIS-3537.

mikeng13: I've deprecated 3537 in favor of 3535. So the requirement from @nchilada for this particular PR is right!

core-api/src/test/java/com/optimizely/ab/event/internal/EventFactoryTest.java
```diff
@@ -249,6 +244,56 @@ public void createImpressionEventIgnoresUnknownAttributes() throws Exception {
...
+
+        EventBatch impression = gson.fromJson(impressionEvent.getBody(), EventBatch.class);
+
+        // verify that no Feature is created for attribute.getKey() -> null
+        for (com.optimizely.ab.event.internal.payload.Attribute feature : impression.getVisitors().get(0).getAttributes()) {
+            assertNotSame(feature.getKey(), attribute.getKey());
+            assertNotSame(feature.getValue(),  null);
+        }
+    }
+
```
nchilada: Can we include explicit unit tests to make sure 
* Integers and Decimals are both tracked. I'm not sure if we can test them distinctly via the compatibility suite, since attribute values have to be serialized via JSON.
* Longs, BigIntegers, and BigDecimals aren't tracked? We definitely can't test these via the compatibility suite.

Maybe test that Booleans are also trackable (in addition to Strings, which I assume are already tested) but I'm less concerned since those can theoretically be covered by the compatibility suite.

core-api/src/main/java/com/optimizely/ab/event/internal/EventFactory.java
```diff
@@ -168,6 +168,15 @@ public LogEvent createConversionEvent(@Nonnull ProjectConfig projectConfig,
         List<Attribute> attributesList = new ArrayList<Attribute>();
 
         for (Map.Entry<String, ?> entry : attributes.entrySet()) {
+            //Ignore null value attribute
+            if (entry.getValue() == null ||
+                    (!(entry.getValue() instanceof String) &&
+                    !(entry.getValue() instanceof Integer) &&
+                    !(entry.getValue() instanceof Double) &&
+                    !(entry.getValue() instanceof Boolean))) {
```
msohailhussain: Just a suggestion, Usually don't like too many Not conditions. so why we can't convert it into one not. 
!(e.getValue() instanceof String || e.getValue() instanceOf Int .... ) should be equal to the above condition.
@nchilada what do u suggest?

nchilada: Seems reasonable to me!

core-api/src/test/java/com/optimizely/ab/event/internal/EventFactoryTest.java
```diff
@@ -249,6 +247,113 @@ public void createImpressionEventIgnoresUnknownAttributes() throws Exception {
...
+        EventBatch impression = gson.fromJson(impressionEvent.getBody(), EventBatch.class);
+
+        // verify that no Feature is created for attribute.getKey() -> null
+        for (com.optimizely.ab.event.internal.payload.Attribute feature : impression.getVisitors().get(0).getAttributes()) {
+            assertNotSame(feature.getKey(), attribute.getKey());
+            assertNotSame(feature.getValue(), null);
+        }
+    }
+
```
nchilada: Should we test attribute selection for _conversion_ events too?

BTW I think we could combine `createImpressionEventIgnoresInvalidAttributes`, `createImpressionEventWithIntegerDecimalBoolAndStringAttributes `, `createImpressionEventIgnoresNullAttributes` - that'd reduce complexity while also confirming that we merely ignore invalid attributes rather than dropping _all_ attributes when one or two are invalid.

core-api/src/main/java/com/optimizely/ab/event/internal/EventFactory.java
```diff
@@ -168,6 +168,16 @@ public LogEvent createConversionEvent(@Nonnull ProjectConfig projectConfig,
         List<Attribute> attributesList = new ArrayList<Attribute>();
 
         for (Map.Entry<String, ?> entry : attributes.entrySet()) {
+            // Filters down to the types of values that are allowed.
```
nchilada: Can we link to https://github.com/optimizely/avro-schemas/blob/baeba19a7bd227aae67026474aec239110c5801d/event_batch/src/main/avro/EventBatch.avsc#L22 since that's the source of truth? Or actually https://developers.optimizely.com/x/events/api/#Attribute_ might be preferable since not everyone can view the avro-schemas repo

core-api/src/test/java/com/optimizely/ab/event/internal/EventFactoryTest.java
mnoman09: Lets leave these Unit Tests. I added one more unit test which is making ConversionEvent and in that unit test we are passing valid and invalid attributes in same attributes map, we are verifying that only valid attributes are getting passed and others are getting ignored. I think that will cover the condition you are asking to verify. So is that ok?

nchilada: Okay, that works too. Thanks!

core-api/src/main/java/com/optimizely/ab/event/internal/EventFactory.java
```diff
@@ -168,6 +168,18 @@ public LogEvent createConversionEvent(@Nonnull ProjectConfig projectConfig,
         List<Attribute> attributesList = new ArrayList<Attribute>();
 
         for (Map.Entry<String, ?> entry : attributes.entrySet()) {
+            // Filter down to the types of values we're allowed to track.
+            // Don't allow Longs, BigIntegers, or BigDecimals - they /can/ theoretically be serialized as JSON numbers
+            // but may take on values that can't be faithfully parsed by the backend.
+            // https://github.com/optimizely/avro-schemas/blob/baeba19a7bd227aae67026474aec239110c5801d/event_batch/src/main/avro/EventBatch.avsc#L22
```
mikeng13: This is a private repo. Please don't mention it here.

mikeng13:
FYI, I ran integration tests for these changes and they passed =)

core-api/src/main/java/com/optimizely/ab/event/internal/EventFactory.java
nchilada: Okay, fair enough. Can we link to https://developers.optimizely.com/x/events/api/#Attribute_ instead?

PR #216, opened 2018-09-21
chore: Prepare release branch for new release

mikeng13:
## Summary
This cherry-picks #214 and #215 into the 2.1.x release branch.

## Tests
The Compat Suite runs on the latest version of the testapp, which is built from the master branch and already has changes to support the 3.0 release. Therefore, this particular release branch compile with that testapp version and we can't run the compat suite tests on it. 

Instead, I've checked out a different branch of the testapp and ran the compat suite tests locally against this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,6 @@ cache:
 branches:
   only:
   - master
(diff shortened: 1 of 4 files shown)


aliabbasrizvi:
Looks like compat failed.

PR #220, opened 2018-09-27
Fix: Allowing Empty UserID as valid field

mnoman09:


core-api/src/main/java/com/optimizely/ab/config/ProjectConfig.java
```diff
@@ -479,7 +479,7 @@ public boolean setForcedVariation(@Nonnull String experimentKey,
         }
 
         // if the user id is invalid, return false.
-        if (userId == null || userId.trim().isEmpty()) {
```
thomaszurkan-optimizely: I think I would like to see some kind of utility isValidUserId.  This allows for a blank or all blanks.  Is that a valid user id (i.e. " ", "     ")?  Maybe || (!userId.isEmpty() && userId.trim().isEmpty())

mnoman09:
@thomaszurkan-optimizely  Yes there are testcases of checking empty user id as valid added in fullstack compatibility suite. @mikeng13 can you please elaborate further.

PR #221, opened 2018-09-27
fix: Removed filtering of unknown attributes and assignment of empty map to null attributes and null event tags.

mnoman09:
Note: first we have to merge #219 PR


core-api/src/main/java/com/optimizely/ab/event/internal/EventFactory.java
```diff
@@ -167,32 +167,34 @@ public LogEvent createConversionEvent(@Nonnull ProjectConfig projectConfig,
...
-            // Don't allow Longs, BigIntegers, or BigDecimals - they /can/ theoretically be serialized as JSON numbers
-            // but may take on values that can't be faithfully parsed by the backend.
-            // https://developers.optimizely.com/x/events/api/#Attribute
-            if (entry.getValue() == null ||
-                    !((entry.getValue() instanceof String) ||
-                    (entry.getValue() instanceof Integer) ||
-                    (entry.getValue() instanceof Double) ||
-                    (entry.getValue() instanceof Boolean))) {
-                continue;
+        if (attributes != null) {
```
thomaszurkan-optimizely: Rather than a large enclosed if != how about if == null continue;  I think it reads better.

msohailhussain:
@thomaszurkan-optimizely We can discuss offline tomorrow, this is the part of audience match type, what is sent to attributes from APIs must be passed same to notification listener. @mikeng13 Tom is right, notification listener has nonnull attributes

core-api/src/main/java/com/optimizely/ab/event/internal/EventFactory.java
msohailhussain: Tom there is no for loop here and if attributes is not null then we have to build attribute list (for loop is inside if) from attributes and later on, we are also building attribute for bot filtering.

mikeng13:
build

mikeng13:
build

mikeng13:
build

PR #224, opened 2018-10-05
fix: Invert the substring match type to match the condition value as a substring of the user attribute value

mikeng13:


mikeng13:
build

PR #225, opened 2018-10-08
Audience combinations

thomaszurkan-optimizely:
with unit tests for parsing audienceConditions.


thomaszurkan-optimizely:
build

core-api/src/main/java/com/optimizely/ab/internal/ConditionUtils.java
```diff
@@ -0,0 +1,171 @@
...
+import com.optimizely.ab.config.audience.UserAttribute;
+import org.json.simple.JSONObject;
+
+import javax.annotation.Nonnull;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+public class ConditionUtils {
+    static public Condition parseConditions(List<Object> rawObjectList) {
```
nchilada: Nice extraction! 🕵

nchilada: Though I still haven't figured out how our JSON parsing works in the Java SDK, so I'll trust @mikeng13 or @aliabbasrizvi to review this part for now

core-api/src/main/java/com/optimizely/ab/internal/ConditionUtils.java
```diff
@@ -0,0 +1,171 @@
...
+                condition = new OrCondition(conditions);
+                break;
+        }
+
+        return condition;
+    }
+
+
+    public static void resolveAudienceIdConditions(@Nonnull ProjectConfig projectConfig,
+                                                   @Nonnull Condition conditions) {
```
nchilada: It's not clear to me what this function does: there's no return value and no obvious side effects. Can you add a JSDoc?

core-api/src/main/java/com/optimizely/ab/internal/ExperimentUtils.java
```diff
@@ -77,4 +85,15 @@ public static boolean isUserInExperiment(@Nonnull ProjectConfig projectConfig,
 
         return false;
     }
+
+    public static Boolean resolveAudienceConditions(@Nonnull ProjectConfig projectConfig,
+                                            @Nonnull Experiment experiment,
+                                            @Nonnull Map<String, ?> attributes) {
+
+        ConditionUtils.resolveAudienceIdConditions(projectConfig, experiment.getAudienceConditions());
```
nchilada: As mentioned above, it's not clear to me what we accomplish by calling `ConditionUtils.resolveAudienceIdConditions`.

core-api/src/main/java/com/optimizely/ab/internal/ExperimentUtils.java
```diff
@@ -59,6 +63,10 @@ public static boolean isExperimentActive(@Nonnull Experiment experiment) {
     public static boolean isUserInExperiment(@Nonnull ProjectConfig projectConfig,
                                              @Nonnull Experiment experiment,
                                              @Nonnull Map<String, ?> attributes) {
+        if (experiment.getAudienceConditions() != null) {
+            Boolean resolveReturn = resolveAudienceConditions(projectConfig, experiment, attributes);
+            return resolveReturn == null ? false : resolveReturn;
+        }
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: I wasn't sure if there would be a case where the audienceConditions was empty but audienceIds were not.

core-api/src/main/java/com/optimizely/ab/config/Experiment.java
```diff
@@ -44,6 +45,7 @@
     private final String groupId;
 
     private final List<String> audienceIds;
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: I am strongly against this idea.  There does not seem to be any logical reason to make a change such as this.

core-api/src/main/java/com/optimizely/ab/config/Experiment.java
```diff
@@ -168,6 +177,7 @@ public String toString() {
                 ", groupId='" + groupId + '\'' +
                 ", status='" + status + '\'' +
                 ", audienceIds=" + audienceIds +
+                ", audienceConditions=" + audienceConditions +
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: logging would be the number one reason. also debugging in the debugger.

core-api/src/main/java/com/optimizely/ab/config/Experiment.java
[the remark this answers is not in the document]
nchilada: This was my reasoning:

> I think this will allow us to
>
> 1. Reduce the amount of forking in SDK code, although this may be irrelevant given how simple `audienceIds` currently is.
> 2. Eventually consolidate on `audienceIds` once all installed SDKs know to treat it as a condition tree. This'd be nice since `audienceConditions` (or whatever other name we might have chosen for the new property) is more confusing.

We should discuss more broadly than in this PR if we want to change the proposed behavior

core-api/src/main/java/com/optimizely/ab/internal/ExperimentUtils.java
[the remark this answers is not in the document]
nchilada: Ah, gotcha. I think if `audienceConditions` are _defined_ but empty, they should still take precedence. They'll only be defined in stale datafiles 👍 

@aliabbasrizvi can you confirm?

core-api/src/main/java/com/optimizely/ab/config/Experiment.java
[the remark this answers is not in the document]
nchilada: Cool, thanks!

core-api/src/main/java/com/optimizely/ab/internal/ConditionUtils.java
```diff
@@ -0,0 +1,207 @@
...
+                break;
+        }
+
+        return condition;
+    }
+
+    /**
+     * resolveAudienceIdConditions walks the conditions list and fills in the audience in any AudienceIdConditions by
+     * looking up audience by id and setting the audience in the AudienceIdCondition.  It uses recursion to walk the list and find
+     * any AudienceIdCondition not set.
```
nchilada: Got it, thanks!

I do think it might be simpler to have `AudienceIdCondition.evaluate` look up its audience by ID, rather than requiring the mapping to have been performed (with its own tree traversal logic) in a priori step. I think the temporal dependency here is kinda spaghetti - it's another way things go wrong and I'm not seeing the benefit (though I might just be missing it)

thomaszurkan-optimizely: I agree that there might be a better way to do this.  We could attempt to resolve during parsing.  This would require that either audiences are parsed first or the resolution happens at project config parse time.  The point is that there is no easy way to get the project config to resolve the audience id at the time of parsing (you are actually in the middle of parsing the project config at this time)

core-api/src/main/java/com/optimizely/ab/config/audience/AudienceIdCondition.java
```diff
@@ -26,24 +27,15 @@
 /**
  * The AudienceIdCondition is a holder for the audience id in
  * {@link com.optimizely.ab.config.Experiment#audienceConditions auienceConditions}.  The AudienceId is later
- * resoloved to Audience before evaluation.  If you do not resolve the AudienceIdCondition before evalutating, the
- * condition will fail.  AudienceIdCondtions are resolved using
- * {@link com.optimizely.ab.internal.ConditionUtils#resolveAudienceIdConditions(com.optimizely.ab.config.ProjectConfig, Condition)}
+ * resolved to Audience before evaluation.  If you do not resolve the AudienceIdCondition before evaluating, the
+ * condition will fail.  AudienceIdConditions are resolved when the ProjectConfig is passed into evaluate.
```
nchilada: I think this JSDoc might be unnecessary now, but if we keep it, I would change:

* "resolved to Audience before evaluation" -> "resolved to Audience _during_ evaluation"
* "If you do not resolve the AudienceIdCondition before evaluating, the condition will fail." -> "If you do not pass the ProjectConfig that is necessary for audience resolution, evaluation will fail."

or something like that.

core-api/src/main/java/com/optimizely/ab/internal/ExperimentUtils.java
```diff
@@ -65,23 +67,32 @@ public static boolean isUserInExperiment(@Nonnull ProjectConfig projectConfig,
...
-            }
+            AudienceIdCondition condition = new AudienceIdCondition(audienceId);
+            conditions.add(condition);
         }
 
-        return false;
+        OrCondition implicitOr = new OrCondition(conditions);
+
+        return implicitOr.evaluate(projectConfig, attributes);
+
```
nchilada: Short and sweet! I guess I'm fine with restricting `experiment.audienceIds` to being an array of strings, but FYI I'm still going to [beat the drum](https://optimizely.atlassian.net/browse/OASIS-3642) to ensure that we infer `"or"` instead of `"not"` in contexts where we're parsing non-leaf nodes in a full condition tree (e.g. `experiment.audienceConditions` and `audience.conditions`). Doing so will help ensure that all our SDKs/snippets behave identically and can't accidentally negate any conditions.

core-api/src/main/java/com/optimizely/ab/internal/ExperimentUtils.java
```diff
@@ -59,22 +63,53 @@ public static boolean isExperimentActive(@Nonnull Experiment experiment) {
...
+            Boolean resolveReturn = resolveAudienceConditions(projectConfig, experiment, attributes);
+            return resolveReturn == null ? false : resolveReturn;
+        }
+        else {
+            Boolean resolveReturn = resolveAudience(projectConfig, experiment, attributes);
+            return resolveReturn == null ? false : resolveReturn;
+        }
+    }
+
+    public static @Nullable Boolean resolveAudience(@Nonnull ProjectConfig projectConfig,
```
nchilada: Maybe rename to `resolveAudienceIds`, for symmetry with `resolveAudienceConditions`? (Or `evaluateAudienceIds` and `evaluateAudienceConditions`, though I don't really care)

core-api/src/main/java/com/optimizely/ab/config/Experiment.java
```diff
@@ -44,6 +45,7 @@
     private final String groupId;
 
     private final List<String> audienceIds;
+    private final Condition audienceConditions;
```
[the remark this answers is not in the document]
nchilada: I still think generics might be worth considering, since they would provide a more rigid model for how we structure `experiment.audienceConditions` vs. `audience.conditions`. This would also be useful if we introduce other, non-audience-related usages of condition trees in the future (as we have in Web).

core-api/src/main/java/com/optimizely/ab/internal/ExperimentUtils.java
nchilada: Oh, I might be missing something, but I'm talking about places like [this](https://github.com/optimizely/java-sdk/blob/da3b24f9c8078722a4bc03cd7fd99b16a1b12552/core-api/src/main/java/com/optimizely/ab/config/parser/JsonSimpleConfigParser.java#L322-L323)

thomaszurkan-optimizely: yes, they are all fixed in this PR.  default is Or

mikeng13:
build

core-api/src/main/java/com/optimizely/ab/internal/ExperimentUtils.java
nchilada: Ah, sorry, I think I'd minimized those diffs! My bad. Thanks so much for doing this - looks good overall, though I have a couple comments which I'll make inline 👍

mikeng13:
build

core-api/src/main/java/com/optimizely/ab/config/parser/ConditionJacksonDeserializer.java
```diff
@@ -0,0 +1,95 @@
...
+        Condition condition;
+        switch (operand) {
+            case "and":
+                condition = new AndCondition(conditions);
+                break;
+            case "or":
+                condition = new OrCondition(conditions);
+                break;
+            case "not": // this makes two assumptions: operator is "not" and conditions is non-empty...
+                condition = new NotCondition(conditions.get(0));
```
nchilada: Can we handle `conditions.length == 0` a little more gracefully, maybe by constructing a `NullCondition`? Merely because of the precedent that we don't enforce any constraints on `audience.conditions` during initialization.

core-api/src/main/java/com/optimizely/ab/config/parser/ConditionJacksonDeserializer.java
```diff
@@ -0,0 +1,95 @@
...
+                break;
+            default:
+                condition = new OrCondition(conditions);
```
nchilada: Important: since `conditionNode.get(0)` wasn't an explicit operator, we should include it as the first operand. Might require a bit of a refactor to make sure it's included in the `conditions` list. Sorry! 🙈

core-api/src/main/java/com/optimizely/ab/config/audience/AudienceIdCondition.java
thomaszurkan-optimizely: fixed

core-api/src/main/java/com/optimizely/ab/internal/ExperimentUtils.java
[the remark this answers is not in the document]
thomaszurkan-optimizely: Ok, if audienceConditions:[] is present in the datafile it will resolve to EmptyCondition which evaluates to true.

core-api/src/main/java/com/optimizely/ab/config/parser/ConditionJacksonDeserializer.java
```diff
@@ -48,20 +48,37 @@ public ConditionJacksonDeserializer() {
...
+            conditionsTree = objectMapper.readTree(conditionsJson);
+        }
         Condition conditions = ConditionJacksonDeserializer.parseConditions(objectMapper, conditionsTree);
 
         return conditions;
     }
 
     protected static Condition parseConditions(ObjectMapper objectMapper, JsonNode conditionNode) throws JsonProcessingException {
+
+        if (conditionNode.size() == 0) {
```
nchilada: Does this identify empty arrays, or empty objects, or both? The former (which could map to an OrCondition with no children) should evaluate to true while the latter (which should map to a UserAttribute) should evaluate to null.

core-api/src/main/java/com/optimizely/ab/config/parser/ConditionJacksonDeserializer.java
```diff
@@ -48,20 +48,37 @@ public ConditionJacksonDeserializer() {
...
+            return new EmptyCondition();
```
nchilada: You could also just construct an OrCondition with no children, since that's semantically what an empty array represents.

(Applies to ConditionUtils.java too)

core-api/src/main/java/com/optimizely/ab/config/parser/ConditionJacksonDeserializer.java
```diff
@@ -81,8 +98,8 @@ else if (subNode.isObject()) {
             case "or":
                 condition = new OrCondition(conditions);
                 break;
-            case "not": // this makes two assumptions: operator is "not" and conditions is non-empty...
-                condition = new NotCondition(conditions.get(0));
+            case "not":
+                condition = new NotCondition(conditions.isEmpty() ? new EmptyCondition() : conditions.get(0));
```
nchilada: `EmptyCondition` evaluates to true, which isn't quite what we want: a `"not"` with no operands is invalid and should evaluate to null.

Perhaps we could define a new `InvalidCondition` that evaluates to true, or construct `new NotCondition(conditions.get(0))` as we were doing before while also updating `NotCondition` to correctly handle a nullable arg.

(Applies to ConditionUtils.java too)

core-api/src/main/java/com/optimizely/ab/config/parser/GsonHelpers.java
```diff
@@ -106,9 +106,13 @@ static Condition parseAudienceConditions(JsonObject experimentJson) {
...
 
-        String stringConditions = experimentJson.get("audienceConditions").getAsString();
+        JsonElement conditionsElement = experimentJson.get("audienceConditions");
 
-        JsonElement conditionsElement = parser.parse(stringConditions);
+        if (!conditionsElement.isJsonArray()) {
+            String stringConditions = experimentJson.get("audienceConditions").getAsString();
+
+            conditionsElement = parser.parse(stringConditions);
+        }
```
nchilada: I don't think we need this fallback at all for `experiment.audienceConditions` - I think it'll **always** be raw JSON. @aliabbasrizvi can you confirm?

core-api/src/main/java/com/optimizely/ab/config/parser/JsonSimpleConfigParser.java
```diff
@@ -155,15 +155,23 @@ public ProjectConfig parseProjectConfig(@Nonnull String json) throws ConfigParse
...
+                        // unable to parse conditions.
+                        Logger.getAnonymousLogger().log(Level.ALL, "problem parsing audience conditions", e);
+                    }
                 }
-                catch (Exception e) {
-                    // unable to parse conditions.
-                    Logger.getAnonymousLogger().log(Level.ALL, "problem parsing audience conditions", e);
+                else {
+                    try {
+                        JSONArray conditionJson = (JSONArray) parser.parse((String)jsonCondition);
```
nchilada: Again, I don't think we need this fallback

core-api/src/main/java/com/optimizely/ab/config/parser/ConditionJacksonDeserializer.java
```diff
@@ -48,20 +48,37 @@ public ConditionJacksonDeserializer() {
...
+        switch (operand) {
+            case "or":
+            case "and":
+            case "not":
+                startingParsingIndex = 1;
+                break;
+            default:
+                operand = "or";
+                break;
+        }
```
nchilada: I like how you extracted an `operand` helper function in the other parser(s)! Maybe do the same here? And maybe even put it in a shared module / define it once, if the types allow?

core-api/src/main/java/com/optimizely/ab/config/parser/ConditionJacksonDeserializer.java
thomaszurkan-optimizely: this would be for empty array only

core-api/src/main/java/com/optimizely/ab/config/parser/ConditionJacksonDeserializer.java
thomaszurkan-optimizely: I don't think this applies anymore.

nchilada: Oh yeah, will resolve. Thanks!

core-api/src/main/java/com/optimizely/ab/config/parser/ConditionJacksonDeserializer.java
```diff
@@ -0,0 +1,129 @@
...
+            startingParsingIndex = 1;
+        }
+
+        for (int i = startingParsingIndex; i < conditionNode.size(); i++) {
+            JsonNode subNode = conditionNode.get(i);
+            if (subNode.isArray()) {
+                conditions.add(ConditionJacksonDeserializer.<T>parseConditions(clazz, objectMapper, subNode));
+            } else if (subNode.isTextual()) {
+                if (clazz != AudienceIdCondition.class) {
+                    throw new InvalidAudienceCondition(String.format("Expected AudienceIdCondition got %s", clazz.getCanonicalName()));
```
nchilada: We say "Expected AudienceIdCondition got SomeOtherType", but I think it would be more accurate to say "Strings not supported for leaves of type SomeOtherType". I.e. we trust the type that we're attempting to parse so we should complain about the datafile instead.

Same thing elsewhere.

core-api/src/main/java/com/optimizely/ab/config/parser/ConditionJacksonDeserializer.java
```diff
@@ -0,0 +1,129 @@
...
+                case "or":
+                case "not":
+                    return op;
+                default:
+                    break;
+            }
+        }
+        return null;
+    }
+    protected static <T> Condition parseConditions(Class<T> clazz, ObjectMapper objectMapper, JsonNode conditionNode)
```
nchilada: Should we construct and return `Condition<T>` instead of `Condition`?

core-api/src/test/java/com/optimizely/ab/config/ProjectConfigTestUtils.java
```diff
@@ -62,14 +63,15 @@ private static ProjectConfig generateValidProjectConfigV2() {
                            ""),
             new Experiment("118", "etag2", "Not started", "2",
                            singletonList("100"),
-                           asList(new Variation("278", "vtag3"),
+                    null,                           asList(new Variation("278", "vtag3"),
```
loganlinn:
```suggestion
                           null,
                           asList(new Variation("278", "vtag3"),
```

core-api/src/test/java/com/optimizely/ab/config/ProjectConfigTestUtils.java
```diff
@@ -62,14 +63,15 @@ private static ProjectConfig generateValidProjectConfigV2() {
...
                                   new Variation("279", "vtag4")),
                            Collections.singletonMap("testUser3", "vtag3"),
                            asList(new TrafficAllocation("278", 4500),
                                   new TrafficAllocation("279", 9000)),
                            ""),
             new Experiment("119", "etag3", "Not started", null,
                            singletonList("100"),
+                    null,
```
loganlinn:
```suggestion
                           null,
```

core-api/src/main/java/com/optimizely/ab/config/audience/match/ExistsMatch.java
```diff
@@ -20,7 +20,7 @@
 
 import javax.annotation.Nullable;
 
-class ExistsMatch extends AttributeMatch<Object> {
+class ExistsMatch implements Match {
```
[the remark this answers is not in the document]
thomaszurkan-optimizely: we don't care about type.  we are matching whether something exists or not.

core-api/src/main/java/com/optimizely/ab/config/parser/ConditionJacksonDeserializer.java
thomaszurkan-optimizely: I do not want to add back in parsing as string after the last PR where you told me that audienceConditions would never be a string.  I'm sticking to that.  Sorry.

core-api/src/main/java/com/optimizely/ab/config/audience/match/ExistsMatch.java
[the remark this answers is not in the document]
nchilada: Okay. It surprised me a little since there is a separate declaration of the `”exists”` match type for each separate dimension type that we might need to support, but I agree that that needn’t influence the SDK implementation 👍

core-api/src/main/java/com/optimizely/ab/config/parser/ConditionJacksonDeserializer.java
nchilada: Okay, if things stay like this, I will need to track this stuff and figure out how to account for SDK differences during datafile build (both right now and in the future). I’ll approve for now but we should discuss this class of SDK differences with the team

core-api/src/main/java/com/optimizely/ab/config/parser/ConditionJacksonDeserializer.java
thomaszurkan-optimizely: But, ["or"] should be null where as [] should be true.

PR #227, opened 2018-10-11
Refact: Copying attributes into new copiedAttributes variable

mnoman09:


core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -818,6 +827,20 @@ public UserProfileService getUserProfileService() {
 
     //======== Helper methods ========//
 
+    /**
+     * Helper function to check that the provided attributes are null if not than it returns a copy
```
mikeng13: This description is wrong =)

PR #228, opened 2018-10-16
fix(NaN/Ifinity): Don't target/track NaN, Infinity, or -Infinity

msohailhussain:


core-api/src/test/java/com/optimizely/ab/config/audience/AudienceConditionEvaluationTest.java
```diff
@@ -228,6 +244,21 @@ public void gtMatchConditionEvaluatesTrue() throws Exception {
...
     }
 
+    /**
+     * Verify that UserAttribute.evaluate for GT match type returns null if the UserAttribute's
+     * value type is invalid number.
+     */
+    @Test
+    public void gtMatchConditionEvaluatesNullWithInvalidAttr() throws Exception {
+        BigInteger bigInteger = new BigInteger("33221312312312312");
+        Double infiniteDouble = 1.2/0;
```
nchilada: Technically this evaluates to `NaN`, not infinity. It's probably cleaner use `Double.NaN`, `Double.POSITIVE_INFINITY`, and `Double.NEGATIVE_INFINITY` where needed

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchType.java
```diff
@@ -72,6 +70,16 @@ else if (conditionValue instanceof Boolean) {
         return new MatchType(matchType, new NullMatch());
     }
 
+    private static boolean isValidNumber(Object conditionValue) {
+        if (conditionValue instanceof Integer || conditionValue instanceof Double) {
```
nchilada: If `conditionValue instanceof Integer`, can we also add a restriction that `Math.abs(value) <= 1e53`? Technically it's unnecessary since _Java_ ints can't be that large, but since this is a reference SDK, it might be helpful to set an example for other SDKs.

nchilada:
@msohailhussain do you know when you might be able to merge this? @mikeng13 it might be nice to get this into the first official 3.0 release.

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchType.java
```diff
@@ -72,6 +70,21 @@ else if (conditionValue instanceof Boolean) {
         return new MatchType(matchType, new NullMatch());
     }
 
+    private static boolean isValidNumber(Object conditionValue) {
+        if (conditionValue instanceof Integer) {
+            if (Math.abs((Integer) conditionValue) > 1e53) {
```
loganlinn:
```suggestion
            return Math.abs((Integer) conditionValue) <= 1e53;
```

PR #230, opened 2018-11-02
fix: created a new matcher for numbers

thomaszurkan-optimizely:
fix(core):Add a new match type for numbers exact match

The object.equals method will use Double.equals and fail if the Number is an Integer.  So, added an explicit ExactNumberMatch to compare actual values.

This fix was relatively small and does not have a large impact.  But, added tests to make sure that "exact" match works for both double and integer.


nchilada:
@thomaszurkan-optimizely looks like this is closed. Should we review this?

thomaszurkan-optimizely:
@nchilada .  Yes please review asap.  I had closed the PR because of the GitHub issues on Friday.

PR #232, opened 2018-11-06
fix: typed audience conditions non-string parsing

thomaszurkan-optimizely:
fix(core):parse audience.conditions as string and typedAudience.conditions as json element (array).

We now parse TypedAudiences slightly different but still store them as Audience types in the ProjectConfig.

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/config/parser/ProjectConfigJacksonDeserializer.java
+++ b/core-api/src/main/java/com/optimizely/ab/config/parser/ProjectConfigJacksonDeserializer.java
@@ -53,16 +53,9 @@ class ProjectConfigJacksonDeserializer extends JsonDeserializer<ProjectConfig> {
             audiences = JacksonHelpers.arrayNodeToList(node.get("audiences"), Audience.class, codec);
         }
 
-        List<TypedAudience> typedAudiencesParsed = null;
-        List<Audience> typedAudiences = null;
+        List<TypedAudience> typedAudiences = null;
         if (node.has("typedAudiences")) {
-            typedAudiencesParsed = JacksonHelpers.arrayNodeToList(node.get("typedAudiences"), TypedAudience.class, codec);
-        }
-        if (typedAudiencesParsed != null) {
-            typedAudiences = new ArrayList<>();
-            for (TypedAudience typedAudience : typedAudiencesParsed) {
-                typedAudiences.add(typedAudience);
-            }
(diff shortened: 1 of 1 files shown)


core-api/src/main/java/com/optimizely/ab/config/parser/AudienceGsonDeserializer.java
```diff
@@ -46,7 +46,7 @@ public Audience deserialize(JsonElement json, Type typeOfT, JsonDeserializationC
         String name = jsonObject.get("name").getAsString();
 
         JsonElement conditionsElement = jsonObject.get("conditions");
-        if (!conditionsElement.isJsonArray()) {
+        if (!typeOfT.toString().contains("TypedAudience")) {
             conditionsElement = parser.parse(jsonObject.get("conditions").getAsString());
```
nchilada: I think `typeOfT == null` allows `typedAudience.conditions` to be stringified and not just raw. I preferred the conditional statement without this bit!

Also, it might be cleaner to structure the above three lines using if-else, or to make the second assignment use the value from the first. It's not immediately obviously that the second assignment is supposed to (unconditionally, IMO) override the first in the case of typed audiences.

```
JsonElement conditionsElement;
if (!typeOfT.toString().contains("TypedAudience")) {
  conditionsElement = jsonObject.get("conditions");
} else {
  conditionsElement = parser.parse(jsonObject.get("conditions").getAsString());
}
```

core-api/src/main/java/com/optimizely/ab/config/parser/AudienceJacksonDeserializer.java
```diff
@@ -58,4 +55,3 @@ public Audience deserialize(JsonParser parser, DeserializationContext context) t
     }
```
nchilada: Can you change
```
JsonNode conditionsJson = node.get("conditions");
if (conditionsJson.isTextual()) {
  conditionsJson = objectMapper.readTree(conditionsJson.textValue());	
}
```
to something like
```
JsonNode conditionsJson = objectMapper.readTree(node.get("conditions").textValue());
```
?

mikeng13:
I am running the audience combo tests locally, will ping here.

nchilada:
> I am running the audience combo tests locally, will ping here.
@mikeng13 audience combinations landed in a different PR, so I think we ought to let this one merge and make any combo-related fixes elsewhere

PR #239, opened 2018-11-21
(fix): added unit tests for leaf node parsing

thomaszurkan-optimizely:
## Summary
- After adding some compatibility tests, we discovered there were some errors in single leaf processing.  Fixed those and added tests.

## Test plan
Added unit tests in all json parsers for single leaf parsing.  
Added a single leaf node in the datafile and have all parsers parse it the same.

## Issues
- Java SDK was failing compatibility tests.


nchilada:
The compat test now passes! Thanks @thomaszurkan-optimizely 😊

PR #240, opened 2018-11-24
update changelog for lastest release. added a bunch of unit tests for…

thomaszurkan-optimizely:
… single leaf parsing

## Summary
- Added single leaf parsing unit tests for all condition parsing types.

## Test plan
Added unit tests for single leaf parsing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,30 @@
 # Optimizely Java X SDK Changelog
 
+## 3.0.0-RC3
+
+November 23rd, 2018
+
+This is the release candidate for the 3.0 SDK, which includes a number of improvements to audience targeting along with a few bug fixes.
+
+### New Features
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,30 @@
 # Optimizely Java X SDK Changelog
 
+## 3.0.0-RC3
+
+November 23rd, 2018
+
+This is the release candidate for the 3.0 SDK, which includes a number of improvements to audience targeting along with a few bug fixes.
+
+### New Features
+* Support for number-valued and boolean-valued attributes. ([#213](https://github.com/optimizely/java-sdk/pull/213))
```
mikeng13: Seems like we are repeating a lot of these bullets. We should probably just include notes for the delta between RC2 and RC3. And do the same for RC1 and RC2. Then we can compile all of it and repeat just once more for 3.0.0.

PR #243, opened 2018-12-06
Fix(Impression event): Resolved method post issue 

mnoman09:
## Summary
- Checked LogEvent request method type if it is get than generate get request and if it is post than generate post request.  

## Issues
- failure asserting results[reach][total_count] == expectation.total_impressions
ERROR:root:expected 1 total visitors; actually found 0 total visitors

### DIFF
--- a/core-httpclient-impl/src/main/java/com/optimizely/ab/event/AsyncEventHandler.java
+++ b/core-httpclient-impl/src/main/java/com/optimizely/ab/event/AsyncEventHandler.java
@@ -177,7 +177,12 @@ public class AsyncEventHandler implements EventHandler {
         @Override
         public void run() {
             try {
-                HttpGet request = generateRequest(logEvent);
+                HttpRequestBase request;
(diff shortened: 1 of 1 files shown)


core-httpclient-impl/src/main/java/com/optimizely/ab/event/AsyncEventHandler.java
```diff
@@ -198,23 +203,31 @@ private HttpGet generateRequest(LogEvent event) throws URISyntaxException {
...
-            int status = response.getStatusLine().getStatusCode();
-            if (status >= 200 && status < 300) {
-                // read the response, so we can close the connection
-                response.getEntity();
-                return null;
-            } else {
-                throw new ClientProtocolException("unexpected response from event endpoint, status: " + status);
-            }
+    @Override
+    public @CheckForNull
```
msohailhussain: Please use in the same line.

PR #249, opened 2018-12-21
fix(number validation): Don't track NaN, Infinity, -Infinity, or > 2^53

mnoman09:
## Summary
- Allow targeting Long and float values
- Don't target Values > 2^53.
- Modified isValidNumber to check if the number is smaller then 2^53 and checking instanceOf float and long attribute value.


core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchType.java
```diff
@@ -72,10 +72,15 @@ public static MatchType getMatchType(String matchType, Object conditionValue) {
...
+            return Math.abs((Integer) conditionValue) <= Math.pow(2, 53);
         } else if (conditionValue instanceof Double) {
             Double value = ((Number) conditionValue).doubleValue();
-            return !(value.isNaN() || value.isInfinite());
+            return !(value.isNaN() || value.isInfinite() || value > Math.pow(2, 53));
+        } else if (conditionValue instanceof Long) {
+            return Math.abs((Long) conditionValue) <= Math.pow(2, 53);
+        } else if (conditionValue instanceof Float) {
+            Float value = ((Number) conditionValue).floatValue();
+            return !(value.isNaN() || value.isInfinite() || value > Math.pow(2, 53));
```
msohailhussain: Both validations look same for double and float, please consolidate into one.

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchType.java
```diff
@@ -72,10 +72,15 @@ public static MatchType getMatchType(String matchType, Object conditionValue) {
 
     private static boolean isValidNumber(Object conditionValue) {
         if (conditionValue instanceof Integer) {
-            return Math.abs((Integer) conditionValue) <= 1e53;
+            return Math.abs((Integer) conditionValue) <= Math.pow(2, 53);
         } else if (conditionValue instanceof Double) {
             Double value = ((Number) conditionValue).doubleValue();
-            return !(value.isNaN() || value.isInfinite());
+            return !(value.isNaN() || value.isInfinite() || value > Math.pow(2, 53));
```
msohailhussain: Math.abs missing or double doesn't take -ve values?

core-api/src/main/java/com/optimizely/ab/event/internal/EventFactory.java
```diff
@@ -218,4 +217,19 @@ public LogEvent createConversionEvent(@Nonnull ProjectConfig projectConfig,
 
         return attributesList;
     }
+
+    private static boolean isValidNumber(Object conditionValue) {
```
msohailhussain: https://github.com/optimizely/java-sdk/pull/249/files#diff-922db348b1c4d7acd0c13f647f829eb6R73

we have two methods with same name, same functionality, can't it go into single helper method.

core-api/src/test/java/com/optimizely/ab/event/internal/EventFactoryTest.java
```diff
@@ -335,6 +335,93 @@ public void createConversionEventIgnoresInvalidAndAcceptsValidAttributes() {
         }
     }
 
+    /**
+     * Verify that passing through an list value attribute causes that attribute to be ignored, rather than
+     * causing an exception to be thrown and passing only the valid attributes.
+     */
+    @Test
+    public void createConversionEventIgnoresInvalidAndAcceptsValidValueOfValidTypeAttributes() {
```
msohailhussain: Please add -ve value cases.

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchType.java
```diff
@@ -72,10 +72,15 @@ public static MatchType getMatchType(String matchType, Object conditionValue) {
 
     private static boolean isValidNumber(Object conditionValue) {
         if (conditionValue instanceof Integer) {
-            return Math.abs((Integer) conditionValue) <= 1e53;
+            return Math.abs((Integer) conditionValue) <= Math.pow(2, 53);
```
msohailhussain: Please add documentation of this method.

core-api/src/main/java/com/optimizely/ab/event/internal/EventFactory.java
msohailhussain: Please add documentation of this method.

core-api/src/main/java/com/optimizely/ab/internal/AttributesUtil.java
```diff
@@ -0,0 +1,23 @@
+package com.optimizely.ab.internal;
```
msohailhussain: Please add header.

core-api/src/test/java/com/optimizely/ab/event/internal/EventFactoryTest.java
msohailhussain: too long method name, please make it bit shorter.

nchilada:
Hi, sorry for the delayed code review. Could we please revive this PR? In addition to merging `master` (which will include changes from #228), I think we need to
* refactor slightly such that we have a single `isValidNumber` implementation for both targeting and tracking
* make sure `isValidNumber` enforces a 2^53 size limits for floating-point numbers and not just for integers. Looks like we missed that in #228!

core-api/src/test/java/com/optimizely/ab/event/internal/EventFactoryTest.java
```diff
@@ -335,6 +335,174 @@ public void createConversionEventIgnoresInvalidAndAcceptsValidAttributes() {
         }
     }
 
+    /**
+     * Verify that passing through an list of invalid value attribute causes that attribute to be ignored, rather than
+     * causing an exception to be thrown and passing only the valid attributes.
+     */
+    @Test
+    public void createConversionEventIgnoresInvalidAcceptValidValOfValidAttr() {
```
aliabbasrizvi: I think you will have to adjust this test to accommodate changes from: https://github.com/optimizely/java-sdk/pull/250

The expectation is that there will be no decisions and `enrich_decisions` should be set to `true`

core-api/src/test/java/com/optimizely/ab/event/internal/EventFactoryTest.java
```diff
@@ -335,6 +335,174 @@ public void createConversionEventIgnoresInvalidAndAcceptsValidAttributes() {
...
+            assertNotSame(feature.getValue(), doubleAttribute);
+        }
+    }
+
+    /**
+     * Verify that passing through an list of -ve invalid attribute value causes that attribute to be ignored, rather than
+     * causing an exception to be thrown and passing only the valid attributes.
+     */
+    @Test
+    public void createConversionEventIgnoresNegativeInvalidAndAcceptsValidValueOfValidTypeAttributes() {
```
aliabbasrizvi: Similar feedback as above.

nchilada:
Thanks for the updates! The new code looks good. But I just realized that when we evaluate targeting conditions, we probably ought to call `isValidNumber` on the user attribute value and not just on the condition value. We're only doing the latter right now.

nchilada:
@mnoman09 do you think you could pull in latest `master`? I think we may have time to approve and merge this now. Sorry for the delays!

core-api/src/test/java/com/optimizely/ab/event/internal/EventFactoryTest.java
```diff
@@ -315,6 +315,148 @@ public void createConversionEventIgnoresInvalidAndAcceptsValidAttributes() {
...
+            attributes,
+            eventTagMap);
+
+        EventBatch impression = gson.fromJson(conversionEvent.getBody(), EventBatch.class);
+
+        //Check valid attributes are getting passed.
+        assertEquals(impression.getVisitors().get(0).getAttributes().get(0).getKey(), boolAttribute.getKey());
+        assertEquals(impression.getVisitors().get(0).getAttributes().get(0).getValue(), validBoolAttribute);
+        assertEquals(impression.getVisitors().get(0).getAttributes().get(1).getKey(), validFloatAttribute.getKey());
+        //In below condition I am checking Value of float with double value because impression gets visitors from json so that converts it into double
```
aliabbasrizvi: nit. Change this to:
```
In the condition below we are checking Value of float with double value because impression gets visitors from JSON so that converts it into double
```

core-api/src/test/java/com/optimizely/ab/event/internal/EventFactoryTest.java
```diff
@@ -315,6 +315,148 @@ public void createConversionEventIgnoresInvalidAndAcceptsValidAttributes() {
...
+
+        LogEvent conversionEvent = factory.createConversionEvent(
+            validProjectConfig,
+            userId,
+            eventType.getId(),
+            eventType.getKey(),
+            attributes,
+            eventTagMap);
+
+        EventBatch impression = gson.fromJson(conversionEvent.getBody(), EventBatch.class);
```
aliabbasrizvi: Why are you calling this impression? Shouldn't this be called conversion?

core-api/src/test/java/com/optimizely/ab/event/internal/EventFactoryTest.java
```diff
@@ -315,6 +315,148 @@ public void createConversionEventIgnoresInvalidAndAcceptsValidAttributes() {
...
+        assumeTrue(datafileVersion >= Integer.parseInt(ProjectConfig.Version.V4.toString()));
+
+        Bucketer mockBucketAlgorithm = mock(Bucketer.class);
+        EventType eventType = validProjectConfig.getEventTypes().get(0);
+
+        List<Experiment> allExperiments = validProjectConfig.getExperiments();
+
+        // Bucket to the first variation for all experiments. However, only a subset of the experiments will actually
+        // call the bucket function.
+        for (Experiment experiment : allExperiments) {
```
aliabbasrizvi: Again. This is not needed right.

core-api/src/test/java/com/optimizely/ab/event/internal/EventFactoryTest.java
```diff
@@ -315,6 +315,148 @@ public void createConversionEventIgnoresInvalidAndAcceptsValidAttributes() {
...
+
+        LogEvent conversionEvent = factory.createConversionEvent(
+            validProjectConfig,
+            userId,
+            eventType.getId(),
+            eventType.getKey(),
+            attributes,
+            eventTagMap);
+
+        EventBatch impression = gson.fromJson(conversionEvent.getBody(), EventBatch.class);
```
aliabbasrizvi: This should technically be called conversion and not impression right?

PR #251, opened 2019-01-02
Feat(AudienceEvalLogging): Added logging in Audience evaluation and its unit tests 

mnoman09:
## Summary
- Added logging in audience evaluation code 
- Added its unit tests to verify logging


core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchType.java
```diff
@@ -64,9 +65,11 @@ public static MatchType getMatchType(String matchType, Object conditionValue) {
                 }
                 break;
             default:
+                logger.warn(String.format("Audience condition \"%s\" uses an unknown match type: %s", userAttribute.toString(), matchType));
                 return new MatchType(matchType, new NullMatch());
         }
 
+        logger.warn(String.format("Audience condition \"%s\" condition value \"%s\" data type is inapplicable", userAttribute.toString(), conditionValue));
```
loganlinn: not justified as `warn` level IMO.

this would be more informative if we logged `matchType`.
also, think we shouldn't be logging `conditionValue` object -- it is potentially very expensive (depends on it's `toString()` impl) and has potential to log undesired data (PII).

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchType.java
```diff
@@ -28,7 +29,7 @@
     private String matchType;
     private Match matcher;
 
-    public static MatchType getMatchType(String matchType, Object conditionValue) {
+    public static MatchType getMatchType(String matchType, Object conditionValue, UserAttribute userAttribute) {
```
loganlinn: :-1: to adding an argument to the signature just for logging

core-api/src/main/java/com/optimizely/ab/internal/ExperimentUtils.java
```diff
@@ -101,9 +105,11 @@ public static boolean isUserInExperiment(@Nonnull ProjectConfig projectConfig,
 
         Condition conditions = experiment.getAudienceConditions();
         if (conditions == null) return null;
-
+        logger.debug(String.format("Evaluating audiences for experiment \"%s\": \"%s\"", experiment.getKey(), conditions.toString()));
         try {
-            return conditions.evaluate(projectConfig, attributes);
+            Boolean result = conditions.evaluate(projectConfig, attributes);
+            logger.info(String.format("Audiences for experiment %s collectively evaluated as " + result, experiment.getKey()));
```
loganlinn: not justified as `info` level IMO.

I would rather leave it up to the user of SDK to log this on the caller side than to put here.

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchType.java
```diff
@@ -64,9 +65,13 @@ public static MatchType getMatchType(String matchType, Object conditionValue) {
                 }
                 break;
             default:
+                logger.error("Audience condition \"{}\" uses an unknown match type: {}", audienceCondition, matchType);
                 return new MatchType(matchType, new NullMatch());
         }
 
+        logger.error("Audience condition \"{}\" condition value data type is inapplicable for match type \"{}\"",
```
msohailhussain: isn't it unreachable code?

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -75,15 +80,31 @@ public Object getValue() {
...
 
         if (!"custom_attribute".equals(type)) {
-            MatchType.logger.error(String.format("condition type not equal to `custom_attribute` %s", type));
+            logger.error("Audience condition \"{}\" has an unknown condition type: {}", this.toString(), type);
             return null; // unknown type
         }
         // check user attribute value is equal
         try {
-            return MatchType.getMatchType(match, value).getMatcher().eval(userAttributeValue);
+            Match matchType = MatchType.getMatchType(match, value, this.toString()).getMatcher();
```
msohailhussain: please see my comments below for this position.

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchType.java
```diff
@@ -28,7 +29,7 @@
     private String matchType;
     private Match matcher;
 
-    public static MatchType getMatchType(String matchType, Object conditionValue) {
+    public static MatchType getMatchType(String matchType, Object conditionValue, String audienceCondition) {
```
msohailhussain: remove arg from here.

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchType.java
msohailhussain: I will recommend to place these logs after this line in UserAttribute. 
Match matchType = MatchType.getMatchType(match, value, this.toString()).getMatcher();
I can't see any difference in line 69 and 75, so we should use both logs there, we may avoid args to add in this method.

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchType.java
msohailhussain: @loganlinn Do you have any suggestion to deal it, where we need to log with userAttributes. I am thinking to add some enum types for NullMatch, when NullMatch is returned, we'll also set NullMatchReasonType in the object. So In two or many more different cases we will check it using ReasonType under NullMatch object.

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchTypeError.java
```diff
@@ -0,0 +1,35 @@
...
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+package com.optimizely.ab.config.audience.match;
+
+public enum MatchTypeError {
+    UNKNOWN_MATCH_TYPE("uses an unknown match type: {}"),
+    INAPPLICABLE_CONDITION_VALUE("condition value data type is inapplicable for match type \"{}\"");
```
nchilada: As suggested in MatchType.java, I don't think we need this enum value

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -76,14 +81,34 @@ public Boolean evaluate(ProjectConfig config, Map<String, ?> attributes) {
...
+            Boolean result = matchType.eval(userAttributeValue);
+
+            if (!(matchType instanceof NullMatch) && result == null) {
+                if (!attributes.containsKey(name)) {
+                    //Missing attribute value
+                    logger.debug("Audience condition \"{}\" evaluated as UNKNOWN because no user attribute for \"{}\" key is passed", this.toString(), name);
+                } else {
+                    //if attribute value is not valid
+                    logger.warn(
+                        "Audience condition \"{}\" evaluated as UNKNOWN because the value for user attribute \"{}\" is inapplicable: \"{}\"",
```
nchilada: We ended up taking y'all's suggestion of logging the name of the offending type! Can we go with
```
"Audience condition \"{}\" evaluated to UNKNOWN because a value of type \"{}\" was passed for user attribute \"{}\""
```
or if we were given a null attribute value:
```
"Audience condition \"{}\" evaluated to UNKNOWN because a null value was passed for the named user attribute"
```
It may be nontrivial to figure out what the type is. i think you may have been discussing with @loganlinn? Let me know if I can help or if we need to change tack.

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -76,14 +81,34 @@ public Boolean evaluate(ProjectConfig config, Map<String, ?> attributes) {
...
+                        this.toString(),
+                        name,
+                        userAttributeValue);
+                }
+            } else if (matchType instanceof NullMatch) {
+                logger.error("Audience condition \"{}\" " + ((NullMatch) matchType).getMatchTypeError().toString(),
```
nchilada: In MatchType.java and MatchTypeError.java, I'm suggesting that we reduce the number of entries in the `MatchTypeError` enum from 2 to 1. Maybe we can get away with inlining that 1 message here instead of defining it in an enum? I'll leave it up to you all.

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -76,14 +81,34 @@ public Boolean evaluate(ProjectConfig config, Map<String, ?> attributes) {
...
+                    this.toString(),
+                    match == null ? "legacy_custom_attribute" : match);
+            }
+            return result;
         } catch (NullPointerException np) {
-            MatchType.logger.error(String.format("attribute or value null for match %s", match != null ? match : "legacy condition"), np);
+            logger.error(String.format("attribute or value null for match %s", match != null ? match : "legacy condition"), np);
```
nchilada: Nit: This is a somewhat cryptic log message. @loganlinn @thomaszurkan-optimizely maybe we can audit our code to include all necessary null checks and then get rid of this `NullPointerException` catcher? Tho maybe it's not worthwhile if we don't expect to get here in practice...

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -75,15 +79,31 @@ public Object getValue() {
...
-            return MatchType.getMatchType(match, value).getMatcher().eval(userAttributeValue);
+            Match matchType = MatchType.getMatchType(match, value, this).getMatcher();
+            Boolean result = matchType.eval(userAttributeValue);
+
+            if (!matchType.getClass().toString().contains("Null") && result == null) {
+                if (!attributes.containsKey(name)) {
+                    //Missing attribute value
+                    logger.debug(String.format("Audience condition \"%s\" evaluated as UNKNOWN because no value was passed for user attribute \"%s\"", this.toString(), name));
+                } else {
+                    logger.warn(
```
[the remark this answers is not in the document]
nchilada: @loganlinn, to clarify, are you saying that this message is redundant with the warning that we would have logged [in castToValueType](https://github.com/optimizely/java-sdk/pull/251/files/5b35599ab7905158a3328d659d79da5b5f02e920#diff-70714a232be8b7ea28ff2448ac70f214)?

core-api/src/main/java/com/optimizely/ab/internal/ExperimentUtils.java
```diff
@@ -73,14 +73,15 @@ public static boolean isUserInExperiment(@Nonnull ProjectConfig projectConfig,
...
-    Boolean evaluateAudience(@Nonnull ProjectConfig projectConfig,
+    @Nullable
+    public static Boolean evaluateAudience(@Nonnull ProjectConfig projectConfig,
                              @Nonnull Experiment experiment,
                              @Nonnull Map<String, ?> attributes) {
         List<String> experimentAudienceIds = experiment.getAudienceIds();
 
         // if there are no audiences, ALL users should be part of the experiment
         if (experimentAudienceIds.isEmpty()) {
+            logger.debug("There is no Audience associated with experiment {}", experiment.getKey());
```
nchilada: Good thinking here. However, I have a couple thoughts

* We should log the `audienceIds` that we use even if (_especially_ if) they're non-empty
* Perhaps we should use the same type of log message that we use for `audienceConditions`:
  ```
  "Evaluating audiences for experiment \"{}\": \"{}\""
  ```
  Example log message:
  ```
  Evaluating audiences for experiment "123": ["456", "789"]
  ```

Reasoning:
* This makes the `audienceIds` -> `audienceConditions` migration more transparent to the customer.
* This may also increase the consistency of our log messages across SDKs. The JavaScript SDK and other dynamic-language SDKs use the same evaluation logic for `audienceIds` and `audienceConditions`.

core-api/src/main/java/com/optimizely/ab/config/audience/match/AttributeMatch.java
```diff
@@ -21,8 +21,8 @@ T castToValueType(Object o, Object value) {
         try {
             if (!o.getClass().isInstance(value) && !(o instanceof Number && value instanceof Number)) {
                 MatchType.logger.warn(
-                    String.format("Incompatible type %s, %s",
-                        o.getClass().getCanonicalName(), value.getClass().getCanonicalName())
+                    "Incompatible type: Attribute value Type {}, Condition value Type {}",
+                    o.getClass().getCanonicalName(), value.getClass().getCanonicalName()
```
nchilada: We should either:
* turn _this_ into the place where we log
  ```
    "Audience condition \"{}\" evaluated to UNKNOWN because a value of type \"{}\" was passed for user attribute \"{}\""
  ```
  , which would require passing down information about the audience condition,
* or we should avoid logging a warning here, instead expecting `UserAttribute.evaluate` to [figure out and log what happened](https://github.com/optimizely/java-sdk/pull/251/files/5b35599ab7905158a3328d659d79da5b5f02e920#diff-289dc99f3875d89845c39fb04ac6b9c1R92).

@loganlinn any thoughts? And my apologies, I might be misunderstanding your original point or missing what's going on with this code.

core-api/src/main/java/com/optimizely/ab/config/audience/AudienceIdCondition.java
```diff
@@ -71,10 +71,14 @@ public Boolean evaluate(ProjectConfig config, Map<String, ?> attributes) {
             audience = config.getAudienceIdMapping().get(audienceId);
         }
         if (audience == null) {
-            logger.error(String.format("Audience not set for audienceConditions %s", audienceId));
+            logger.error("Audience not set for audienceConditions {}", audienceId);
             return null;
         }
-        return audience.getConditions().evaluate(config, attributes);
+        logger.debug("Starting to evaluate audience {} with conditions: \"{}\"", audience.getName(), audience.getConditions());
```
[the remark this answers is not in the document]
mnoman09: This is how its getting logged you can check unit tests where I have verified.
 Example: "Starting to evaluate audience not_firefox_users with conditions: "[and, [or, [not, [or, {name='browser_type', type='custom_attribute', match='null', value='firefox'}]]]]"

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchType.java
```diff
@@ -64,10 +64,10 @@ public static MatchType getMatchType(String matchType, Object conditionValue) {
                 }
                 break;
             default:
-                return new MatchType(matchType, new NullMatch());
+                return new MatchType(matchType, new NullMatch(MatchTypeError.UNKNOWN_MATCH_TYPE));
         }
 
-        return new MatchType(matchType, new NullMatch());
+        return new MatchType(matchType, new NullMatch(MatchTypeError.INAPPLICABLE_CONDITION_VALUE));
```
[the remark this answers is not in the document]
mnoman09: This condition will occur when condition value type do not meet match type requirement.

core-api/src/main/java/com/optimizely/ab/config/audience/AudienceIdCondition.java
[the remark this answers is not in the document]
nchilada: Ooh, yeah, you're right. It looks like we achieve that via `AndCondition.toString`, `OrCondition.toString`, and `NotCondition.toString`. I thought there'd be an abstract/interface method stemming from the `Condition` interface, but it makes sense to use `Object.toString` instead.

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -81,7 +81,7 @@ public Boolean evaluate(ProjectConfig config, Map<String, ?> attributes) {
         Object userAttributeValue = attributes.get(name);
 
         if (!"custom_attribute".equals(type)) {
-            logger.error("Audience condition \"{}\" has an unknown condition type: {}", this.toString(), type);
+            logger.warn("Audience condition \"{}\" has an unknown condition type. You may need to upgrade to a newer release of the Optimizely SDK", this.toString(), type);
```
nchilada: No need to pass `type` now.

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchType.java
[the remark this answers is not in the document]
nchilada: Oh, I see - we get here if we enter one of the _non-default_ cases for match type and fail the check for the condition value type. Makes sense!

I may suggest a change to the `INAPPLICABLE_CONDITION_VALUE` message.

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchTypeError.java
nchilada: [Explained in MatchType.java](https://github.com/optimizely/java-sdk/pull/251#discussion_r247898163)

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchTypeError.java
```diff
@@ -0,0 +1,35 @@
...
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+package com.optimizely.ab.config.audience.match;
+
+public enum MatchTypeError {
+    UNKNOWN_MATCH_TYPE("uses an unknown match type. You may need to upgrade to a newer release of the Optimizely SDK"),
+    INAPPLICABLE_CONDITION_VALUE("has an unknown condition value type. You may need to upgrade to a newer release of the Optimizely SDK");
```
nchilada: Thanks for explaining what `INAPPLICABLE_CONDITION_VALUE` is for! Can we change the message a little bit to clarify? I think this would be sufficient:
```
"has an unexpected value type. You may need to upgrade to a newer release of the Optimizely SDK"
```
After thinking about it for a while, I guess it's okay to suggest that the customer upgrade their SDK here. It could help if we eventually support non-primitive values for `exact` match, or lexicographic `gt`/`lt` semantics for strings values, etc. 🤷

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
nchilada: [Explained in MatchType.java](https://github.com/optimizely/java-sdk/pull/251#discussion_r247898163)

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchType.java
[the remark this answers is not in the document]
nchilada: Oh, I see - we get here if we enter one of the _non-default_ cases for match type and fail the check for the condition value type. Makes sense!

I may suggest a change to the `INAPPLICABLE_CONDITION_VALUE` message.

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchTypeError.java
```diff
@@ -18,8 +18,8 @@
 package com.optimizely.ab.config.audience.match;
 
 public enum MatchTypeError {
-    UNKNOWN_MATCH_TYPE("uses an unknown match type: {}"),
-    INAPPLICABLE_CONDITION_VALUE("condition value data type is inapplicable for match type \"{}\"");
+    UNKNOWN_MATCH_TYPE("uses an unknown match type. You may need to upgrade to a newer release of the Optimizely SDK"),
+    INAPPLICABLE_CONDITION_VALUE("has an unknown condition value type. You may need to upgrade to a newer release of the Optimizely SDK");
```
nchilada: Thanks for explaining what `INAPPLICABLE_CONDITION_VALUE` is for! Can we change the message a little bit to clarify? I think this would be sufficient:
```
"has an unexpected value type. You may need to upgrade to a newer release of the Optimizely SDK"
```
After thinking about it for a while, I guess it's okay to suggest that the customer upgrade their SDK here. It could help if we eventually support non-primitive values for `exact` match, or lexicographic `gt`/`lt` semantics for strings values, etc. 🤷

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
nchilada: No need to pass `type` now.

core-api/src/main/java/com/optimizely/ab/config/audience/match/AttributeMatch.java
nchilada: I'm starting to lean towards the second approach. It's definitely hacky (since the code that does the logging is attempting to figure out what went wrong in the code that evaluates the audience condition) but I like all the [customer-friendly log messages](https://github.com/optimizely/java-sdk/pull/251/files/5957ef142dcc245090c466b2dcd1f384f5584b5e#diff-289dc99f3875d89845c39fb04ac6b9c1R92) that we now have in `UserAttribute.evaluate`

nchilada: As of [commit 2be5a31](https://github.com/optimizely/java-sdk/pull/251/commits/2be5a3104b2e5b171f9ca07f353aa6366f514c5a), it looks like we're going with the second approach. Thanks for the updates! I'll resolve this conversation.

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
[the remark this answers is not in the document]
nchilada: As of [commit 2be5a31](https://github.com/optimizely/java-sdk/pull/251/commits/2be5a3104b2e5b171f9ca07f353aa6366f514c5a), we're no longer logging a warning in `castToValueType`

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchType.java
nchilada: > not justified as `warn` level IMO.

I dunno, a customer probably won't expect an audience condition to become inactive for this reason, and therefore probably ought to be warned. If it helps, the updated log message suggests that the customer upgrade their SDK.

> this would be more informative if we logged `matchType`.

It's included (along with `type`, `name`, and `value`) in our logging of the audience condition (the `userAttribute` variable).

> also, think we shouldn't be logging `conditionValue` object -- it is potentially very expensive (depends on it's `toString()` impl) and has potential to log undesired data (PII).

Good feedback. We're not logging it any more!

core-api/src/test/java/com/optimizely/ab/config/audience/AudienceConditionEvaluationTest.java
```diff
@@ -108,6 +114,65 @@ public void invalidMatchCondition() throws Exception {
...
+
+    /**
+     * Verify that UserAttribute.evaluate returns null on invalid attribute type.
+     */
+    @Test
+    public void unexpectedAttributeTypeNull() throws Exception {
+        UserAttribute testInstance = new UserAttribute("browser_type", "custom_attribute", "gt", 20);
+        assertNull(testInstance.evaluate(null, Collections.singletonMap("browser_type", null)));
+        logbackVerifier.expectMessage(Level.ERROR, String.format("Cannot evaluate targeting condition since the value for attribute is an incompatible type"));
+        logbackVerifier.expectMessage(Level.ERROR, String.format("Greater than match failed"));
```
nchilada: Ah, I know that these errors were already being logged before the current PR, but we don't want them any longer since we only want warnings instead. Can you remove the error logging from AttributeMatch.java, SubstringMatch.java, ExactNumberMatch.java, GTMatch.java, and LTMatch.java?

Thanks for including these assertion in the initial tests, however! I wouldn't have noticed otherwise. 😬

core-api/src/test/java/com/optimizely/ab/config/audience/AudienceConditionEvaluationTest.java
```diff
@@ -108,6 +114,65 @@ public void invalidMatchCondition() throws Exception {
...
+        logbackVerifier.expectMessage(Level.WARN, String.format("Audience condition \"%s\" evaluated to UNKNOWN because a null value was passed for user attribute \"browser_type\"", testInstance.toString()));
```
nchilada: Nit: Inline the audience condition

core-api/src/test/java/com/optimizely/ab/config/audience/AudienceConditionEvaluationTest.java
```diff
@@ -108,6 +114,65 @@ public void invalidMatchCondition() throws Exception {
...
+    /**
+     * Verify that UserAttribute.evaluate returns null on missing attribute value.
+     */
+    @Test
+    public void missingAttribute() throws Exception {
+        UserAttribute testInstance = new UserAttribute("browser_type", "custom_attribute", "gt", 20);
+        assertNull(testInstance.evaluate(null, Collections.EMPTY_MAP));
+        logbackVerifier.expectMessage(Level.ERROR, String.format("Cannot evaluate targeting condition since the value for attribute is an incompatible type"));
+        logbackVerifier.expectMessage(Level.ERROR, String.format("Greater than match failed"));
+        logbackVerifier.expectMessage(Level.DEBUG, String.format("Audience condition \"%s\" evaluated to UNKNOWN because no value was passed for user attribute \"%s\"", testInstance.toString(), testInstance.getName()));
```
nchilada: Nit: Inline the audience condition and attribute name

core-api/src/test/java/com/optimizely/ab/config/audience/AudienceConditionEvaluationTest.java
```diff
@@ -108,6 +114,65 @@ public void invalidMatchCondition() throws Exception {
...
+    /**
+     * Verify that UserAttribute.evaluate returns null on passing null attribute object.
+     */
+    @Test
+    public void nullAttribute() throws Exception {
+        UserAttribute testInstance = new UserAttribute("browser_type", "custom_attribute", "gt", 20);
+        assertNull(testInstance.evaluate(null, null));
+        logbackVerifier.expectMessage(Level.ERROR, String.format("Cannot evaluate targeting condition since the value for attribute is an incompatible type"));
+        logbackVerifier.expectMessage(Level.ERROR, String.format("Greater than match failed"));
+        logbackVerifier.expectMessage(Level.DEBUG, String.format("Audience condition \"%s\" evaluated to UNKNOWN because no value was passed for user attribute \"%s\"", testInstance.toString(), testInstance.getName()));
```
nchilada: Nit: Inline the audience condition and attribute name

core-api/src/test/java/com/optimizely/ab/config/audience/AudienceConditionEvaluationTest.java
```diff
@@ -108,6 +114,65 @@ public void invalidMatchCondition() throws Exception {
...
+    }
+
+    /**
+     * Verify that UserAttribute.evaluate returns null on unknown condition type.
+     */
+    @Test
+    public void unknownConditionType() throws Exception {
+        UserAttribute testInstance = new UserAttribute("browser_type", "blah", "exists", "firefox");
+        assertNull(testInstance.evaluate(null, testUserAttributes));
+        logbackVerifier.expectMessage(Level.WARN, String.format("Audience condition \"%s\" has an unknown condition type. You may need to upgrade to a newer release of the Optimizely SDK", testInstance.toString()));
```
nchilada: Nit: Inline the audience condition

core-api/src/test/java/com/optimizely/ab/internal/ExperimentUtilsTest.java
```diff
@@ -130,8 +140,49 @@ public void isUserInExperimentReturnsTrueIfExperimentHasNoAudiences() {
...
     public void isUserInExperimentEvaluatesEvenIfExperimentHasAudiencesButUserHasNoAttributes() {
         Experiment experiment = projectConfig.getExperiments().get(0);
+        Audience audience = projectConfig.getAudience(experiment.getAudienceIds().get(0));
+        Boolean result = isUserInExperiment(projectConfig, experiment, Collections.<String, String>emptyMap());
+        assertTrue(result);
+        logbackVerifier.expectMessage(Level.DEBUG, String.format("Evaluating audiences for experiment \"%s\": \"[100]\"", experiment.getKey()));
+        logbackVerifier.expectMessage(Level.DEBUG, String.format("Starting to evaluate audience %s with conditions: \"%s\"", audience.getName(), audience.getConditions()));
+        logbackVerifier.expectMessage(Level.ERROR, String.format("Cannot evaluate targeting condition since the value for attribute is an incompatible type"));
+        logbackVerifier.expectMessage(Level.INFO, String.format("Audience %s evaluated to %s", audience.getName(), result));
+        logbackVerifier.expectMessage(Level.INFO, String.format("Audiences for experiment %s collectively evaluated to %s", experiment.getKey(), result));
```
nchilada: Nit: Inline all the variables into the expectation strings. Same thing below

core-api/src/test/java/com/optimizely/ab/config/audience/AudienceConditionEvaluationTest.java
```diff
@@ -113,8 +113,11 @@ public void invalidMatchCondition() throws Exception {
     @Test
     public void invalidMatch() throws Exception {
         UserAttribute testInstance = new UserAttribute("browser_type", "custom_attribute", "blah", "chrome");
+        String expectedAudience = testInstance.toString();
         assertNull(testInstance.evaluate(null, testUserAttributes));
-        logbackVerifier.expectMessage(Level.WARN, String.format("Audience condition \"%s\" uses an unknown match type. You may need to upgrade to a newer release of the Optimizely SDK", testInstance.toString()));
+        logbackVerifier.expectMessage(Level.WARN,
+            String.format("Audience condition \"%s\" uses an unknown match type. You may need to upgrade to a newer release of the Optimizely SDK",
+                expectedAudience));
```
nchilada: Hmm, by "inline the audience condition" I meant can we do something like
```java
logbackVerifier.expectMessage(Level.WARN, "Audience condition {name:'blah', type='custom_attribute', match='blah', value='chrome'} uses an unknown match type. You may need to upgrade to a newer release of the Optimizely SDK");
```
This way, if `UserAttribute.toString` behaves unexpectedly, we can tell by looking at our test cases.

(Same thing everywhere below)

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -76,14 +81,40 @@ public Boolean evaluate(ProjectConfig config, Map<String, ?> attributes) {
         Object userAttributeValue = attributes.get(name);
 
         if (!"custom_attribute".equals(type)) {
-            MatchType.logger.error(String.format("condition type not equal to `custom_attribute` %s", type));
+            logger.warn("Audience condition \"{}\" has an unknown condition type. You may need to upgrade to a newer release of the Optimizely SDK", this.toString());
```
loganlinn: calling `this.toString()` is unnecessary as the logger will do this if you just pass `this`, but it will do it lazily in case `WARN` messages are not output

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -76,14 +81,40 @@ public Boolean evaluate(ProjectConfig config, Map<String, ?> attributes) {
...
             return null; // unknown type
         }
         // check user attribute value is equal
         try {
-            return MatchType.getMatchType(match, value).getMatcher().eval(userAttributeValue);
+            Match matchType = MatchType.getMatchType(match, value).getMatcher();
+            Boolean result = matchType.eval(userAttributeValue);
+
+            if (!(matchType instanceof NullMatch) && result == null) {
```
loganlinn: I'm opposed to this approach. The `instanceof` and numerous logic branches are hard to follow and logging makes assumptions about the implementation details of matcher. This may be correct right now, but not guaranteed in the future. It smells out of being out of place and not a proper utilization of `MatchType`/`Matcher`/etc

nchilada: Okay, smells a little funky to me too.

The alternative is logging from the matcher itself. Given the type of log message that we want, that means passing down the entire audience condition to the matcher, which I think you'd [been unsure about](https://github.com/optimizely/java-sdk/pull/251#discussion_r244823028). And we'd also want to pass an additional boolean if we want to preserve distinct messages for "no value" and "a null value".

How should we proceed?

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
nchilada: Ooh, yeah, that makes sense! +1 for only passing `this`

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchTypeError.java
```diff
@@ -0,0 +1,35 @@
...
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+package com.optimizely.ab.config.audience.match;
+
+public enum MatchTypeError {
```
thomaszurkan-optimizely: Should we call this NullMatchTypeErrors since they are only used by NullMatch?

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -76,14 +81,40 @@ public Boolean evaluate(ProjectConfig config, Map<String, ?> attributes) {
...
+                    }
+                }
+            } else if (matchType instanceof NullMatch) {
+                logger.warn("Audience condition \"{}\" " + ((NullMatch) matchType).getMatchTypeError().toString(),
+                    this.toString());
+            }
+            return result;
         } catch (NullPointerException np) {
-            MatchType.logger.error(String.format("attribute or value null for match %s", match != null ? match : "legacy condition"), np);
+            logger.error(String.format("attribute or value null for match %s", match != null ? match : "legacy condition"), np);
```
thomaszurkan-optimizely: Can we remove the String.format?

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchTypeError.java
nchilada: 💯

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
nchilada: My opinion (which I will also mention in other SDK's PRs) are that it's okay to pass additional data for logging purposes. @msohailhussain @mikeng13 FYI

thomaszurkan-optimizely: Sorry, I'm so late to the party.  But, I agree: It stinks.  2 questions on approach:
1.  What's wrong with logging the null messages on the match type and remove the if instanceof?
2.  Why not just create two exceptions and catch for those and log that way?  It is way cleaner than this.

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -76,14 +81,40 @@ public Boolean evaluate(ProjectConfig config, Map<String, ?> attributes) {
...
+                            this,
+                            name);
+                    }
+                }
+            } else if (matchType instanceof NullMatch) {
+                logger.warn("Audience condition \"{}\" " + ((NullMatch) matchType).getNullMatchTypeErrors().toString(),
+                    this);
+            }
+            return result;
         } catch (NullPointerException np) {
```
thomaszurkan-optimizely: Add a couple of new exception type instead of ```matchType instanceof NullMatch```?

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -76,14 +81,41 @@ public Boolean evaluate(ProjectConfig config, Map<String, ?> attributes) {
...
+                            name);
+                    }
+                }
+            }
+            return result;
         } catch (NullPointerException np) {
-            MatchType.logger.error(String.format("attribute or value null for match %s", match != null ? match : "legacy condition"), np);
+            logger.error("attribute or value null for match {}", match != null ? match : "legacy condition", np);
+            return null;
+        } catch (NullMatch nm) {
```
thomaszurkan-optimizely: This should be an exception.  There should be one per error.

core-api/src/main/java/com/optimizely/ab/config/audience/match/NullMatchTypeErrors.java
```diff
@@ -0,0 +1,35 @@
...
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+package com.optimizely.ab.config.audience.match;
+
+public enum NullMatchTypeErrors {
```
thomaszurkan-optimizely: get rid of this class

thomaszurkan-optimizely: create two exceptions UnknownMatchTypeException and UnexpectedValueTypeException.

core-api/src/main/java/com/optimizely/ab/config/audience/match/NullMatch.java
```diff
@@ -16,20 +16,14 @@
...
 
-import javax.annotation.Nullable;
-
-class NullMatch extends AttributeMatch<Object> {
-    @SuppressFBWarnings("URF_UNREAD_FIELD")
-    Object value;
-
-    protected NullMatch() {
-        this.value = null;
+    protected NullMatch(NullMatchTypeErrors nullMatchTypeErrors) {
```
thomaszurkan-optimizely: get rid of this constructor.

core-api/src/main/java/com/optimizely/ab/config/audience/match/NullMatch.java
```diff
@@ -16,20 +16,14 @@
  */
 package com.optimizely.ab.config.audience.match;
 
-import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+public class NullMatch extends Throwable {
```
thomaszurkan-optimizely: create two new exceptions with the suffix Exception and set the appropriate errorDescription

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -76,16 +82,42 @@ public Boolean evaluate(ProjectConfig config, Map<String, ?> attributes) {
...
         // check user attribute value is equal
         try {
-            return MatchType.getMatchType(match, value).getMatcher().eval(userAttributeValue);
+            Match matchType = MatchType.getMatchType(match, value).getMatcher();
+            Boolean result = matchType.eval(userAttributeValue);
+
+            if (result == null) {
+                if (!attributes.containsKey(name)) {
+                    //Missing attribute value
+                    logger.debug("Audience condition \"{}\" evaluated to UNKNOWN because no value was passed for user attribute \"{}\"", this, name);
```
msohailhussain: a NULL value was passed, ```not no value```

PR #259, opened 2019-01-28
Rework attribute audience conditions

loganlinn:
## Summary

Addresses concerns of complexity in business logic and logging
strategy for audience conditions on user attributes while maintaining the
same matching behavior. Only a minor changes have been made some of the
log messages, but should not impact the ability for SDK end-user to
debug.

The `Match` abstraction and `MatchType` class hindered the ability to add detailed logging in a non-intrusive manner (#251).
The `Match` interface and implementations were replaced by classes that
implement `Condition` interface directly and share a common logical workflow.

### Additional Notes
-   Introduces `AbstractAttributeCondition` for the common logic and flow shared between all match operations.
-   Replaces the `castToValueType` method that performed unnecessary reflection and was difficult to reason about -- the class that is required for a match operation is known ahead of time.
-   Generalize string-based conditions (`StringAttributeCondition` and `SubstringAttributeCondition`) to support any `CharSequence`; not just `String`.
-   Avoid special-case logic when `UserAttribute` is not valid by evaluating with dummy `VoidCondition` that always evaluates to `null`.
-   More efficient `toString()` implementations using `StringBuilder` (generated by IntelliJ IDEA "StringBuilder (JDK 1.5)" template)
-   Improve error messages for failed assertions made by `LogbackVerifier` by displaying the expected log message that was never observed.
-   Further improvements can be made to this code, such as removing
object allocations when condition is evaluated and cleaning up the
associated `ConditionUtils.parseConditions` logic, but that is left
intact to minimize amount of change. I added `TODO(llinn)`
comments where I saw where future improvements could be made.

## Test plan
- Existing and unit tests
- Added `UserAttributeTest` unit test


core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -20,27 +20,33 @@
...
+ * Represents a leaf node in audience condition AST.
+ *
+ * Defines a matching operation to evaluate on a specific attribute.
+ * The behavior of {@link #evaluate(ProjectConfig, Map)} is determined by specified
+ * match operation.
+ *
+ * If the given match type and value are not compatible, i.e. a substring match with non-string value,
+ * this class will always evaluate to UNKNOWN by returning {@code null}.
+ *
+ * TODO(llinn) convert this to a POJO and separate from {@link Condition} evaluation
```
nchilada: Not following POJO in this context. Do you mean, convert it to be encapsulated by a leaf node rather than directly acting as the leaf node? Or something like that?

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -75,66 +81,55 @@ public Object getValue() {
...
-                    }
-                }
-            }
-            return result;
-        } catch (UnknownMatchTypeException | UnexpectedValueTypeException ex) {
-            logger.warn("Audience condition \"{}\" " + ex.getMessage(),
-                this);
-        } catch (NullPointerException np) {
-            logger.error("attribute or value null for match {}", match != null ? match : "legacy condition", np);
+            condition = ConditionUtils.leafCondition(type, match, name, value);
```
nchilada: Semantically, I'm not following why a UserAttribute (which is a Condition) should need to go and get a Condition. Are we preparing for other condition types (e.g. a DeviceTypeCondition which would also implement Condition) that might want to share the same match types?

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
```diff
@@ -75,66 +81,55 @@ public Object getValue() {
...
+        } catch (InvalidAudienceCondition e) {
+            logger.warn(
+                "Audience condition \"{}\" is invalid: {}. You may need to upgrade to a newer release of the Optimizely SDK",
```
nchilada: Can we log a more specific error message here, per the [spec](https://docs.google.com/document/d/1DfZzY-dReyu-hElaPngS849C6HC1wFXzuC_Ir0ShCdo/edit#heading=h.y3umnfybr1lu)? Maybe we could embed the details in `e`.

core-api/src/main/java/com/optimizely/ab/internal/VoidCondition.java
```diff
@@ -14,13 +14,23 @@
...
-    private static String message = "uses an unknown match type. You may need to upgrade to a newer release of the Optimizely SDK";
+import javax.annotation.Nullable;
+import java.util.Map;
 
-    public UnknownMatchTypeException() {
-        super(message);
+/**
+ * Dummy condition that always evaluates to {@code null}
+ */
+public final class VoidCondition<T> implements Condition<T> {
```
nchilada: How does this differ from `NullCondition`?

core-api/src/main/java/com/optimizely/ab/internal/VoidCondition.java
```diff
@@ -14,13 +14,23 @@
...
+    static final VoidCondition INSTANCE = new VoidCondition();
```
nchilada: Maybe we can have different versions of VoidCondition that log [different warning messages](https://docs.google.com/document/d/1DfZzY-dReyu-hElaPngS849C6HC1wFXzuC_Ir0ShCdo/edit#heading=h.gc2x9d7t8f88) when they're evaluated?

core-api/src/main/java/com/optimizely/ab/config/audience/match/AbstractAttributeCondition.java
```diff
@@ -0,0 +1,134 @@
...
+ * <ol>
+ *   <li>extracts the untyped attribute value to evaluate</li>
+ *   <li>filter the attribute to types supported by match logic</li>
+ *   <li>evaluate whether typed input is a match</li>
+ * </ol>
+ *
+ * @param <T> type of value that input is matched against
+ * @param <U> type of converted input that is required to match
+ */
+public abstract class AbstractAttributeCondition<T, U> implements Condition<T> {
```
nchilada: Semantically (and maybe even architecturally 😕) it's not clear to me how this differs from UserAttribute.

core-api/src/main/java/com/optimizely/ab/config/audience/match/EqualsAttributeCondition.java
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * Matches input using standard {@link Object#equals(Object)}.
+ *
+ * If an attribute is set to value that is not an instance of {@code inputClass},
+ * this will always evaluate to UNKNOWN.
+ *
+ * @param <T> type of value to match input against
+ */
+public class EqualsAttributeCondition<T, U> extends AbstractAttributeCondition<T, U> {
```
nchilada: It's surprising that a class with this name is only used for exact _boolean_ conditions, and not for exact number or exact string conditions.

core-api/src/main/java/com/optimizely/ab/config/audience/match/StringAttributeCondition.java
```diff
@@ -0,0 +1,63 @@
...
+package com.optimizely.ab.config.audience.match;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Objects;
+
+/**
+ * Matches content of {@link String} against any {@link CharSequence} input.
+ */
+public class StringAttributeCondition extends AbstractAttributeCondition<String, CharSequence> {
```
nchilada: It's surprising that a class with this name is only used for _exact_ string conditions and not for substring conditions.

core-api/src/main/java/com/optimizely/ab/config/audience/match/NumericAttributeCondition.java
```diff
@@ -0,0 +1,120 @@
...
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.Objects;
+
+/**
+ * Abstract condition for {@link Number} comparisons.
+ * Comparisons are made by {@link Double#compare(double, double)}.
+ */
+public abstract class NumericAttributeCondition extends AbstractAttributeCondition<Number, Number> {
```
nchilada: It's interesting to have a class that handles multiple match types for a particular condition value type. Other SDKs don't really do this; they always branch first on the match type and then on the value type. Though this isn't a huge concern for me; like you, I'd rather figure out the class hierarchy and see what falls out of that.

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
aliabbasrizvi: @nchilada you feel `e.getMessage` doesn't have enough details?

core-api/src/test/java/com/optimizely/ab/internal/LogbackVerifier.java
```diff
@@ -84,6 +84,11 @@ private void verify() throws Throwable {
                 public boolean matches(final Object argument) {
                     return expectedEvent.matches((ILoggingEvent) argument);
                 }
+
+                @Override
+                public void describeTo(org.hamcrest.Description description) {
```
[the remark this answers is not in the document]
mikecdavis: This combined with the `toString()` below allows for human readable differences to be displayed when the `ArgumentMatcher` fails to match.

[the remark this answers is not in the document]
mikecdavis: That being said I'm not a fan of this style of testing anyway and would be more than happy to remove this class entirely :)

mikeng13:
CLosing this due to inactivity

PR #270, opened 2019-03-08
(feat): Feature Management getFeatureVariable Listener

mnoman09:
## Summary
- Introduction of OnDecision Notification Listener.
- Returning default feature variable value if featureEnabled is false.
- Implemented onDecisionListener in:
- get_feature_variable_string
- get_feature_variable_boolean
- get_feature_variable_integer
- get_feature_variable_double

##Ticket: 
[OASIS-4228](https://optimizely.atlassian.net/projects/OASIS/issues/OASIS-4228)


aliabbasrizvi:
> Returning default feature variable value if featureEnabled is false.
Isn't this already the case?

mnoman09:
> > Returning default feature variable value if featureEnabled is false.
> 
> Isn't this already the case?

No it was not the case already, we did that change in all sdks

core-api/src/main/java/com/optimizely/ab/notification/DecisionInfoEnums.java
```diff
@@ -0,0 +1,41 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *
+ * See the License for the specific language governing permissions and      *
+ * limitations under the License.                                           *
+ ***************************************************************************/
+
+package com.optimizely.ab.notification;
+
+public class DecisionInfoEnums {
+    public enum GetFeatureVariableDecisionInfo {
+        SOURCE_EXPERIMENT_KEY("source_experiment_key"),
```
aliabbasrizvi: nit. I think alphabetizing will be a good order for this.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -649,24 +650,77 @@ String getFeatureVariableValueForType(@Nonnull String featureKey,
...
             }
+            featureEnabled = featureDecision.variation.getFeatureEnabled();
         } else {
             logger.info("User \"{}\" was not bucketed into any variation for feature flag \"{}\". " +
                     "The default value \"{}\" for \"{}\" is being returned.",
                 userId, featureKey, variableValue, variableKey
             );
         }
 
+        Object convertedValue = convertStringToType(variableValue, variableType);
```
mikecdavis: This conversion is now happening twice per api call. Once here and once in the originating method.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -649,24 +650,77 @@ String getFeatureVariableValueForType(@Nonnull String featureKey,
...
+        Map<String, Object> decisionInfo = new HashMap<>();
+        decisionInfo.put(DecisionInfoEnums.FeatureVariableDecisionInfo.FEATURE_KEY.toString(), featureKey);
```
mikecdavis: Not a valid use of Enums. What you're doing here just requires constants or can even be a builder.
Via constants this would reduce to:
```
decisionInfo.put(DECISION_FEATURE_KEY, featureKey);
decisionInfo.put(DECISION_FEATURE_ENABLED, featureEnabled);
...
```
Via a builder it would be:
```
Map<String, Object> decisionInfo = DecisionInfoMap.builder()
    .withFeatureKey(featureKey)
    .withFeatureEnabled(featureEnabled)
    ...
    .build();
```

I personally like the builder approach since within the builder you'd abstract some of the conditional logic below.
Actually, in the context of the notification listener the entire thing can benefit from the builder pattern which actually transforms into a fluent pattern if you in corporate the notification itself.

```
DecisionNotification.builder()
    .withUserId(userId)
    .withAttributes(attributes)
    .withFeatureKey(featureKey)
    .withFeatureEnabled(featureEnabled)
    ...
    .send();
```
A much cleaner interface :)

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -649,24 +650,77 @@ String getFeatureVariableValueForType(@Nonnull String featureKey,
...
+            NotificationCenter.DecisionNotificationType.FEATURE_VARIABLE.toString(),
+            userId,
+            copiedAttributes,
+            decisionInfo);
+
         return variableValue;
     }
 
+    // Helper method which takes type and variable value and convert it to object to use in Listener DecisionInfo object variable value
+    private Object convertStringToType(String variableValue, FeatureVariable.VariableType type) {
```
mikecdavis: Refactor so we're not doing String to Type conversion multiple times and in multiple places.

core-api/src/main/java/com/optimizely/ab/notification/DecisionInfoEnums.java
```diff
@@ -0,0 +1,41 @@
...
+ * Unless required by applicable law or agreed to in writing, software      *
+ * distributed under the License is distributed on an "AS IS" BASIS,        *
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. *
+ * See the License for the specific language governing permissions and      *
+ * limitations under the License.                                           *
+ ***************************************************************************/
+
+package com.optimizely.ab.notification;
+
+public class DecisionInfoEnums {
```
mikecdavis: Remove Enum class in favor of something more appropriate. See previous comment about constants vs builder vs fluent approaches.

core-api/src/main/java/com/optimizely/ab/notification/DecisionNotificationListener.java
```diff
@@ -0,0 +1,53 @@
...
+        } else {
+            attributes = new HashMap<>();
+        }
+        assert (args[3] instanceof java.util.Map);
+        Map<String, ?> decisionInfo = (Map<String, ?>) args[3];
+        onDecision(type, userId, attributes, decisionInfo);
+    }
+
+    @Override
+    public abstract void onDecision(@Nonnull String type,
```
mikecdavis: Would require a larger refactor, but we can generalize much better if we used actual notification Objects as opposed to positional arguments which as we can see in this example can lead to bugs since one can easily swap type for userId and attributes for decisionInfo. In this example if we had a `DecisionNotification` object with members `type`, `userId` etc, then everything becomes much more explicit.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -445,16 +446,19 @@ public Boolean getFeatureVariableBoolean(@Nonnull String featureKey,
...
+                userId,
+                attributes,
+                FeatureVariable.VariableType.BOOLEAN
+            );
+            return variableValue;
+        } catch (Exception exception) {
+            logger.error("BooleanTypeException while trying to parse \"" + variableValue +
+                "\" as Boolean. " + exception);
         }
         return null;
```
aliabbasrizvi: You can just return `variableValue` here

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -495,20 +499,21 @@ public Double getFeatureVariableDouble(@Nonnull String featureKey,
...
+        Double variableValue = null;
+        try {
+            variableValue = getFeatureVariableValueForType(
+                featureKey,
+                variableKey,
+                userId,
+                attributes,
+                FeatureVariable.VariableType.DOUBLE
+            );
+            if (variableValue != null) {
```
aliabbasrizvi: This check seems redundant.  You can just change last line to be `return variableValue`

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -514,20 +510,19 @@ public Double getFeatureVariableDouble(@Nonnull String featureKey,
...
+                featureKey,
+                variableKey,
+                userId,
+                attributes,
+                FeatureVariable.VariableType.DOUBLE
+            );
+            return variableValue;
+        } catch (Exception exception) {
+            logger.error("NumberFormatException while trying to parse \"" + variableValue +
+                "\" as Double. " + exception);
```
loganlinn: @mnoman09 this and a [other](https://github.com/optimizely/java-sdk/pull/270/files#diff-9719ba1e8c37710e7b92b94f0330e552R577) [log](https://github.com/optimizely/java-sdk/pull/270/files#diff-9719ba1e8c37710e7b92b94f0330e552R729) statements in this PR append the exception as a string on log message. the logger should be passed the exception object as a last argument

PR #271, opened 2019-03-08
(feat): Feature Management isFeatureEnabled Listener

mnoman09:
## Summary
- Introduction of OnDecision Notification Listener.
- Returning default feature variable value if featureEnabled is false.
- Implemented onDecisionListener and added unit tests in:
- is_feature_enabled
- get_enabled_features

##Ticket: 
[OASIS-4228](https://optimizely.atlassian.net/projects/OASIS/issues/OASIS-4228)


core-api/src/main/java/com/optimizely/ab/notification/ActivateNotificationListener.java
```diff
@@ -24,7 +24,6 @@
 import javax.annotation.Nonnull;
 import java.util.Map;
 
-
```
aliabbasrizvi: Seems like an unnecessary change.

core-api/src/main/java/com/optimizely/ab/notification/DecisionInfoEnums.java
```diff
@@ -0,0 +1,39 @@
...
+        FeatureDecisionInfo(String key) {
+            this.key = key;
+        }
+
+        @Override
+        public String toString() {
+            return key;
+        }
+    }
+
```
aliabbasrizvi: nit. remove extraneous line

core-api/src/main/java/com/optimizely/ab/notification/DecisionNotificationListenerInterface.java
```diff
@@ -0,0 +1,36 @@
...
+import javax.annotation.Nonnull;
+import java.util.Map;
+
+public interface DecisionNotificationListenerInterface {
+
+    /**
+     * onDecision called when an activate was triggered
+     *
+     * @param type         - The notification type.
+     * @param userId       - The userId passed into activate.
```
aliabbasrizvi: Please update the docstrings here as mentioned in the other PR.

core-api/src/test/java/com/optimizely/ab/OptimizelyTest.java
```diff
@@ -2636,6 +2638,277 @@ public void getVariationWithInvalidDatafile() throws Exception {
...
+                }
+            });
+
+        ArrayList<String> featureFlags = (ArrayList<String>) optimizely.getEnabledFeatures(testUserId,
+            new HashMap<String, String>());
+        assertEquals(2, featureFlags.size());
+
+        verify(mockEventHandler, times(2)).dispatchEvent(any(LogEvent.class));
+
+        assertTrue(optimizely.notificationCenter.removeNotificationListener(notificationId));
```
aliabbasrizvi: Where is the check for notification listener being called with correct params happening?

core-api/src/main/java/com/optimizely/ab/notification/DecisionNotificationListener.java
```diff
@@ -0,0 +1,54 @@
...
+public abstract class DecisionNotificationListener implements NotificationListener, DecisionNotificationListenerInterface {
+
+    /**
+     * Base notify called with var args.  This method parses the parameters and calls the abstract method.
+     *
+     * @param args - variable argument list based on the type of notification.
+     */
+    @Override
+    public final void notify(Object... args) {
+        assert (args[0] instanceof String);
```
mnoman09: @aliabbasrizvi here we added multiple assert checks which will throw exception if invalid arguments are passed and it will make sure that correct arguments are passed. so I think we dont need to add this in unit tests

aliabbasrizvi: We should have type specific checks i.e. for type: experiment we should check `decisionInfo` has correct experiment and variation key. Similarly for type: feature we should check `decisionInfo` has correct data.

Can you point me to where such assertions are happening? If not, then we have incomplete tests.

core-api/src/test/java/com/optimizely/ab/OptimizelyTest.java
```diff
@@ -2636,6 +2638,309 @@ public void getVariationWithInvalidDatafile() throws Exception {
 
     //======== Notification listeners ========//
 
+    boolean isListenerCalled = false;
```
mikecdavis: This is not a safe way to determine if the listener got called. If we wanted to have a class method then it's lifecycle should be handled via `@Before` and `@After` methods. But in this case we should add a counter to the `DecisionNotificationListener` implementation. Ideally, in the form of an `AtomicInteger` that initializes to zero and increments every time `onDecision` is called. Then within each test we instantiate an instance of this listener, then assert on the value of the counter.

core-api/src/test/java/com/optimizely/ab/OptimizelyTest.java
```diff
@@ -2785,6 +2783,294 @@ public void activateUserNotInAudienceWithListener() throws Exception {
...
+        final Map<String, String> testUserAttributes = new HashMap<>();
+        testUserAttributes.put(ATTRIBUTE_HOUSE_KEY, AUDIENCE_GRYFFINDOR_VALUE);
+        testUserAttributes.put(testBucketingIdKey, testBucketingId);
+
+        final Map<String, Object> testDecisionInfoMap = new HashMap<>();
+        testDecisionInfoMap.put(VARIATION_KEY, null);
+        testDecisionInfoMap.put(FEATURE_KEY, validFeatureKey);
+        testDecisionInfoMap.put(FEATURE_ENABLED, true);
+        testDecisionInfoMap.put(SOURCE, FeatureDecision.DecisionSource.ROLLOUT);
+            testDecisionInfoMap.put(SOURCE_INFO, new HashMap<>());
```
aliabbasrizvi: nit. fix indentation

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -397,18 +401,32 @@ public Boolean isFeatureEnabled(@Nonnull String featureKey,
...
         }
 
+        DecisionNotification decisionNotification = DecisionNotification.newFeatureDecisionNotificationBuilder()
+            .withUserId(userId)
+            .withAttributes(copiedAttributes)
+            .withFeatureKey(featureKey)
+            .withFeatureEnabled(featureEnabled)
+            .withSource(decisionSource)
+            .withExperimentKey(sourceExperimentKey)
+            .withVariationKey(sourceVariationKey)
```
aliabbasrizvi: May be remove these 2 build options and replace with `withSourceInfo` option which takes `SourceInfo` object.

You can then define 2 implementations `SourceInfo`. 1 for rollout and 1 for feature test and then use a get method on them to get the source info that you build in the builder here. That may be a better way to structure this.

mnoman09: Are you suggesting to make methods of get source info in optimizely or in DecisionNotification.java? 
By the way I think current implementation is much simpler and fully utilizing builder pattern.

aliabbasrizvi: I am suggesting something like:
```
public interface SourceInfo {
  HashMap<String, Object> get()
}

public class FeatureTestSourceInfo implements SourceInfo {
}

public class RolloutSourceInfo implements SourceInfo {
}
```

This may allow us to hopefully add more `SourceInfo` options in the future.

aliabbasrizvi: Discussed offline. This will happen in a follow up.

PR #273, opened 2019-03-14
Fix test isolation in OptimizelyTest class.

mikecdavis:
## Summary
- Simplify parameterization of OptimizelyTest
- Build ProjectConfig directly from the datafile JSON to ensure isolation between tests.
- Update individual tests that were reliant on previous test state.

These changes fix the non-determinism between test cases within OptimizelyTest. This was caused by passing singleton instances of ProjectConfig to the parameterized test runner, which were being mutated via calls to `setForcedVariation` in certain test cases.

## Test plan
This is a pure refactor to improve test repeatability. No new tests were added, but existing tests were updated to reflect their expected state.

### DIFF
--- a/core-api/src/test/java/com/optimizely/ab/OptimizelyTest.java
+++ b/core-api/src/test/java/com/optimizely/ab/OptimizelyTest.java
@@ -199,7 +199,7 @@ public class OptimizelyTest {
         logbackVerifier.expectMessage(Level.DEBUG, "Dispatching impression event to URL test_url with params " +
             testParams + " and payload \"{}\"");
 
-        // HACK to fix test. Not sure if the logging varifier is even neccesary.
+        // Force variation to null to get expected log output.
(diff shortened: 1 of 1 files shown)


core-api/src/test/java/com/optimizely/ab/OptimizelyTest.java
```diff
@@ -91,25 +91,16 @@
...
-                validProjectConfigV2(),
-                noAudienceProjectConfigV2()
             },
             {
-                3,
                 validConfigJsonV3(),
-                noAudienceProjectConfigJsonV3(),
-                validProjectConfigV3(),
-                noAudienceProjectConfigV3()
+                noAudienceProjectConfigJsonV3(),  // FIX-ME this is not a valid v3 datafile
```
aliabbasrizvi: I see it is a v2 file. 😞

PR #275, opened 2019-03-22
Refact: Added builder pattern in decision listener

mnoman09:
## Summary
- Refactored FeatureVariable Decision listener PR and applied builder pattern 
- Updated notification listener to directly send notification from it and added its method 

Once these changes will get approved than we can move forward and we can apply these changes in all decision listener PRs


core-api/src/main/java/com/optimizely/ab/notification/NotificationCenter.java
```diff
@@ -194,6 +204,13 @@ public int addNotificationListener(NotificationType notificationType, Notificati
      * @return true if removed otherwise false (if the notification is already registered, it returns false).
      */
     public boolean removeNotificationListener(int notificationID) {
+        for (NotificationHolder holder : decisionListenerHolder) {
```
msohailhussain: @mikecdavis What do you suggest here? There is separate logic to remove for DecisionListener object.

core-api/src/main/java/com/optimizely/ab/notification/DecisionNotificationListener.java
```diff
@@ -16,38 +16,20 @@
...
-        Map<String, ?> decisionInfo = (Map<String, ?>) args[3];
-        onDecision(type, userId, attributes, decisionInfo);
-    }
-
-    @Override
-    public abstract void onDecision(@Nonnull String type,
-                                    @Nonnull String userId,
-                                    @Nonnull Map<String, ?> attributes,
-                                    @Nonnull Map<String, ?> decisionInfo);
+    void onDecision(@Nonnull DecisionNotification decisionNotification);
```
mikecdavis: I've been trying to see if we can use generics to maintain some higher order generality. Ideally NotificationLister originally would have been defined like:
```
public interface NotificationListener<T> {
    void notify(T notification);
}
```
but since it wasn't trying to think the best way forward.

mnoman09: Yes, you are right we can use generics but I think than we have to refactor Track and Activate listeners as well. If I am understanding this correctly?

core-api/src/main/java/com/optimizely/ab/notification/DecisionNotification.java
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        public DecisionNotification build() {
+            decisionInfo = new HashMap<>();
+            decisionInfo.put(FEATURE_KEY, featureKey);
+            decisionInfo.put(FEATURE_ENABLED, featureEnabled);
+            decisionInfo.put(VARIABLE_KEY, variableKey);
+            decisionInfo.put(VARIABLE_TYPE, variableType);
+            decisionInfo.put(VARIABLE_VALUE, variableValue);
+            if (featureDecision.decisionSource != null && featureDecision.decisionSource.equals(FeatureDecision.DecisionSource.EXPERIMENT)) {
```
loganlinn:
```suggestion
            if (featureDecision != null && FeatureDecision.DecisionSource.EXPERIMENT.equals(featureDecision.decisionSource)) {
```

core-api/src/main/java/com/optimizely/ab/notification/DecisionNotification.java
```diff
@@ -0,0 +1,167 @@
...
+            this.attributes = attributes;
+            return this;
+        }
+
+        public FeatureVariableDecisionNotificationBuilder withFeatureKey(String featureKey) {
+            this.featureKey = featureKey;
+            return this;
+        }
+
+        public FeatureVariableDecisionNotificationBuilder withFeatureEnabled(Boolean featureEnabled) {
```
loganlinn:
```suggestion
        public FeatureVariableDecisionNotificationBuilder withFeatureEnabled(boolean featureEnabled) {
```

core-api/src/main/java/com/optimizely/ab/notification/DecisionNotification.java
```diff
@@ -0,0 +1,167 @@
...
+                decisionInfo.put(SOURCE_EXPERIMENT_KEY, featureDecision.experiment.getKey());
+                decisionInfo.put(SOURCE_VARIATION_KEY, featureDecision.variation.getKey());
+                decisionInfo.put(SOURCE, featureDecision.decisionSource);
+            } else {
+                decisionInfo.put(SOURCE_EXPERIMENT_KEY, null);
+                decisionInfo.put(SOURCE_VARIATION_KEY, null);
+                decisionInfo.put(SOURCE, FeatureDecision.DecisionSource.ROLLOUT);
+            }
+
+            this.type = NotificationCenter.DecisionNotificationType.FEATURE_VARIABLE.toString();
```
loganlinn: If `type` is not something that's settable, it's more clear if it's set in the constructor. Maybe it should even be parameter of `BaseDecisionNotificationBuilder` constructor.

core-api/src/main/java/com/optimizely/ab/notification/DecisionNotification.java
```diff
@@ -0,0 +1,167 @@
...
+
+            return super.build();
```
loganlinn: It doesn't seem like the builder class is buying us much and it makes this code harder to follow (stateful logic & having to jump between two classes). Would be more straightforward to just instantiate here

```java
            return new DecisionNotification(
                type,
                userId,
                attributes,
                decisionInfo);
```

Regarding the defaulting of `attributes = new HashMap<>()`, that can be done at field declaration on this class or in the `DecisionNotification` constructor.

core-api/src/main/java/com/optimizely/ab/notification/DecisionNotification.java
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    protected DecisionNotification(String type,
+                                   String userId,
+                                   Map<String, ?> attributes,
+                                   Map<String, ?> decisionInfo) {
+        this.type = type;
+        this.userId = userId;
+        this.attributes = attributes;
+        this.decisionInfo = decisionInfo;
```
loganlinn: Are all of these fields allowed to be null? I see in `BaseDecisionNotificationBuilder`, that we default `attributes` to empty map -- that should be moved here (otherwise assert non-null in constructor) to make it consistent, otherwise nulls can slip in.

core-api/src/main/java/com/optimizely/ab/notification/DecisionNotification.java
```diff
@@ -0,0 +1,167 @@
...
+                type,
+                userId,
+                attributes,
+                decisionInfo);
+        }
+    }
+
+    public static class FeatureVariableDecisionNotificationBuilder extends BaseDecisionNotificationBuilder {
+
+        public static String FEATURE_KEY = "feature_key";
```
loganlinn:
```suggestion
        public static final String FEATURE_KEY = "feature_key";
```

loganlinn: All of these constants should be marked `final`

core-api/src/main/java/com/optimizely/ab/notification/NotificationCenter.java
```diff
@@ -96,30 +102,34 @@ public Class getNotificationTypeClass() {
     public NotificationCenter() {
         notificationsListeners.put(NotificationType.Activate, new ArrayList<NotificationHolder>());
         notificationsListeners.put(NotificationType.Track, new ArrayList<NotificationHolder>());
-        notificationsListeners.put(NotificationType.Decision, new ArrayList<NotificationHolder>());
     }
 
     // private list of notification by notification type.
     // we used a list so that notification order can mean something.
     private Map<NotificationType, ArrayList<NotificationHolder>> notificationsListeners = new HashMap<NotificationType, ArrayList<NotificationHolder>>();
+    private ArrayList<NotificationHolder> decisionListenerHolder = new ArrayList<>();
```
loganlinn:
```suggestion
    private List<NotificationHolder> decisionListenerHolder = new ArrayList<>();
```

core-api/src/main/java/com/optimizely/ab/notification/NotificationCenter.java
```diff
@@ -96,30 +102,34 @@ public Class getNotificationTypeClass() {
...
-                @Override
-                public void onDecision(@Nonnull String type, @Nonnull String userId, @Nonnull Map<String, ?> attributes, @Nonnull Map<String, ?> decisionInfo) {
-                    decisionNotificationListenerInterface.onDecision(type, userId, attributes, decisionInfo);
+    public int addDecisionNotificationListener(DecisionNotificationListener decisionNotificationListener) {
+        int id = this.notificationListenerID.incrementAndGet();
+        if (decisionNotificationListener instanceof DecisionNotificationListener) {
+            for (NotificationHolder holder : decisionListenerHolder) {
+                if (holder.decisionNotificationListener == decisionNotificationListener) {
+                    logger.warn("Notification listener was already added");
+                    return -1;
```
loganlinn: Should it return the ID of the existing listener?

I don't think `warn` is necessary. Is there any potential harm in attempting to add the same listener again? Seems OK to me if idempotent.

core-api/src/main/java/com/optimizely/ab/notification/NotificationCenter.java
```diff
@@ -96,30 +102,34 @@ public Class getNotificationTypeClass() {
...
-    public int addDecisionNotificationListener(final DecisionNotificationListenerInterface decisionNotificationListenerInterface) {
-        if (decisionNotificationListenerInterface instanceof DecisionNotificationListener) {
-            return addNotificationListener(NotificationType.Decision, (NotificationListener) decisionNotificationListenerInterface);
-        } else {
-            return addNotificationListener(NotificationType.Decision, new DecisionNotificationListener() {
-                @Override
-                public void onDecision(@Nonnull String type, @Nonnull String userId, @Nonnull Map<String, ?> attributes, @Nonnull Map<String, ?> decisionInfo) {
-                    decisionNotificationListenerInterface.onDecision(type, userId, attributes, decisionInfo);
+    public int addDecisionNotificationListener(DecisionNotificationListener decisionNotificationListener) {
+        int id = this.notificationListenerID.incrementAndGet();
```
loganlinn: Consider moving this call to where `id` is actually needed. Since there are early-returns in this method, we could bump the id sequence unnecessarily.

core-api/src/main/java/com/optimizely/ab/notification/NotificationCenter.java
```diff
@@ -176,12 +186,12 @@ public int addNotificationListener(NotificationType notificationType, Notificati
         }
 
         for (NotificationHolder holder : notificationsListeners.get(notificationType)) {
-            if (holder.notificationListener == notificationListener ) {
+            if (holder.notificationListener == notificationListener) {
                 logger.warn("Notification listener was already added");
                 return -1;
```
loganlinn: I see that returning `-1` is existing behavior. It's misleading and worth changing IMO. Maybe a later PR, but maybe add a TODO

core-api/src/main/java/com/optimizely/ab/notification/NotificationCenter.java
```diff
@@ -194,25 +204,46 @@ public int addNotificationListener(NotificationType notificationType, Notificati
...
-        for (NotificationType type : NotificationType.values()) {
-            for (NotificationHolder holder : notificationsListeners.get(type)) {
-                if (holder.notificationId == notificationID) {
-                    notificationsListeners.get(type).remove(holder);
-                    logger.info("Notification listener removed {}", notificationID);
-                    return true;
-                }
+
+        for (ArrayList<NotificationHolder> notificationHolders : notificationsListeners.values()) {
+            if(removeNotificationListener(notificationID, notificationHolders)) {
```
loganlinn:
```suggestion
            if (removeNotificationListener(notificationID, notificationHolders)) {
```

core-api/src/main/java/com/optimizely/ab/notification/NotificationCenter.java
```diff
@@ -194,25 +204,46 @@ public int addNotificationListener(NotificationType notificationType, Notificati
...
     public boolean removeNotificationListener(int notificationID) {
-        for (NotificationType type : NotificationType.values()) {
-            for (NotificationHolder holder : notificationsListeners.get(type)) {
-                if (holder.notificationId == notificationID) {
-                    notificationsListeners.get(type).remove(holder);
-                    logger.info("Notification listener removed {}", notificationID);
-                    return true;
-                }
+
+        for (ArrayList<NotificationHolder> notificationHolders : notificationsListeners.values()) {
```
loganlinn:
```suggestion
        for (List<NotificationHolder> notificationHolders : notificationsListeners.values()) {
```

core-api/src/main/java/com/optimizely/ab/notification/NotificationCenter.java
```diff
@@ -194,25 +204,46 @@ public int addNotificationListener(NotificationType notificationType, Notificati
...
     }
 
+    /**
+     * Helper method to iterate find NotificationHolder in an List identified by the notificationId
+     *
+     * @param notificationID the id passed back from add notification.
+     * @param notificationHolderList list from which to remove notification listener.
+     * @return true if removed otherwise false
+     */
+    private boolean removeNotificationListener(int notificationID, ArrayList<NotificationHolder> notificationHolderList) {
```
loganlinn:
```suggestion
    private boolean removeNotificationListener(int notificationID, List<NotificationHolder> notificationHolderList) {
```

loganlinn: side-note: a `Map`, like `LinkedHashMap`, keyed by `id` would make things easier but also retain ordering...

core-api/src/main/java/com/optimizely/ab/notification/DecisionNotification.java
```diff
@@ -149,7 +141,7 @@ public DecisionNotification build() {
             decisionInfo.put(VARIABLE_KEY, variableKey);
             decisionInfo.put(VARIABLE_TYPE, variableType);
             decisionInfo.put(VARIABLE_VALUE, variableValue);
-            if (featureDecision.decisionSource != null && featureDecision.decisionSource.equals(FeatureDecision.DecisionSource.EXPERIMENT)) {
+            if (featureDecision.decisionSource != null && FeatureDecision.DecisionSource.EXPERIMENT.equals(featureDecision.decisionSource)) {
```
loganlinn: Potential `NullPointerException` if `featureDecision` is null

core-api/src/main/java/com/optimizely/ab/notification/DecisionNotification.java
```diff
@@ -0,0 +1,161 @@
...
+
+    public Map<String, ?> getDecisionInfo() {
+        return decisionInfo;
+    }
+
+    public static FeatureVariableDecisionNotificationBuilder newFeatureVariableBuilder() {
+        return new FeatureVariableDecisionNotificationBuilder();
+    }
+
+    private static class BaseDecisionNotificationBuilder {
```
mikecdavis: This base class is not needed.

aliabbasrizvi:
@loganlinn can you take another look?

PR #287, opened 2019-04-25
Implement default ProjectConfigManager.

mikecdavis:
## Summary
- Extract ProjectConfig interface and create DatafileProjectConfig.
- Create ProjectConfigManager and refactor for dynamic ProjectConfig updates.
- Add PollingProjectConfigManager implementation.
- Add Apache HttpClient implementation of ProjectConfigManager.
- Move forced variations mapping into DecisionService.
- Add ability to block execution until ProjectConfig resolves.


core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -867,20 +848,27 @@ public boolean setForcedVariation(@Nonnull String experimentKey,
...
+        }
+
+        return decisionService.getForcedVariation(experiment, userId);
     }
 
     /**
      * @return the current {@link ProjectConfig} instance.
      */
-    @Nonnull
+    @Nullable
```
mikeng13: I think this represents a breaking change right? This will cause compilation failures when the user updates to the latest 3.x.x version

core-api/src/main/java/com/optimizely/ab/UnknownExperimentException.java
```diff
@@ -16,8 +16,8 @@
  */
 package com.optimizely.ab;
 
-import com.optimizely.ab.config.Experiment;
 import com.optimizely.ab.config.ProjectConfig;
+import com.optimizely.ab.config.Experiment;
```
mikeng13: Please update the header for this file.

core-api/src/main/java/com/optimizely/ab/UnknownEventTypeException.java
```diff
@@ -16,8 +16,8 @@
  */
 package com.optimizely.ab;
 
-import com.optimizely.ab.config.EventType;
 import com.optimizely.ab.config.ProjectConfig;
+import com.optimizely.ab.config.EventType;
```
mikeng13: Please update the header for this file.

core-api/src/test/java/com/optimizely/ab/config/DatafileProjectConfigTest.java
```diff
@@ -0,0 +1,173 @@
+/**
+ *
+ *    Copyright 2016-2019, Optimizely and contributors
```
mikeng13: Should just be 2019

java-quickstart/src/main/java/com/optimizely/Example.java
```diff
@@ -1,65 +1,76 @@
+/****************************************************************************
+ * Copyright 2016-2019, Optimizely, Inc. and contributors                   *
```
mikeng13: I think this was added in 2018

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -79,16 +74,17 @@
...
-    ProjectConfig projectConfig;
-    @VisibleForTesting
     final EventHandler eventHandler;
     @VisibleForTesting
     final ErrorHandler errorHandler;
-    private boolean isValid;
+
+    private final ProjectConfigManager projectConfigManager;
+
+    // TODO should be private
```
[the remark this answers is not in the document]
mikecdavis: I added a getter in https://github.com/optimizely/java-sdk/pull/288. Providing access to member variables in this way is not best practice. I can remove the comment can save for other PR if you'd rather.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
mikecdavis: It's not a compilation failure, but more of a hint to your IDE and static code analysis since it can't be fully respected at runtime. The reason I changed it is because it is nullable today even before this change :) Marking as `@Nonnull` can lead to uncaught NPEs.

For example trace [Optimizely#initialize](https://github.com/optimizely/java-sdk/blob/master/core-api/src/main/java/com/optimizely/ab/Optimizely.java#L114) to see how this can be null better yet check out this [test](https://github.com/optimizely/java-sdk/blob/master/core-api/src/test/java/com/optimizely/ab/OptimizelyTest.java#L1567) to see what would be returned.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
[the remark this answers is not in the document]
mikeng13: Nah, this is fine, just wanted to know the context behind this.

core-api/src/main/java/com/optimizely/ab/config/PollingProjectConfigManager.java
```diff
@@ -82,21 +89,25 @@ void setConfig(ProjectConfig projectConfig) {
...
-     * TODO add timeout to we don't hang indefinitely..
      */
     @Override
     public ProjectConfig getConfig() {
         if (started) {
             try {
-                completableFuture.get();
-            } catch (InterruptedException | ExecutionException e) {
-                logger.warn("Interrupted waiting for valid ProjectConfig");
+                boolean aquired = countDownLatch.await(blockingTimeoutPeriod, blockingTimeoutUnit);
```
mikeng13: typo: `acquired` :)

PR #288, opened 2019-04-30
Refactor notification center to use generics.

mikecdavis:
## Summary
- Define new `Notification` interface to mark supported notification models.
- Define new generic `NotificationHandler<T>` interface
- Define `NotifcationManager<T>` class to own specific class of notifications.
- Refactor NotificationCenter to hold a map of `NotificationManager` instances.
- Define explicit `ActivateNotification` class 
- Refactor `ActivateNotificationListener` to implement `NotificationHandler<ActivateNotification>`
- Define explicit `TrackNotification` class 
- Refactor `TrackNotificationListener` to implement `NotificationHandler<TrackNotification>`

The current patterns surrounding the `NotificationCente`r did not extend well and relied on `Object...` array arguments into the `NotificationListener` interface. This resulted in the need for dedicated abstract classes, enums and helper methods to be implemented for each notification. Moving to a generic `NotificationHandler<T>` interface and `NotifcationManager<T>` means that new notifications can be defined simply by implementing a new `Notification` model which is type safe,  discoverable and self describing.

## Test plan
Unit and integration testing along with the compatibility suite.


mikecdavis:
This is failing compatibility-suite changes since I removed the explicit DecisionNotificationListener class and methods for a more extensible pattern. This should be resolved with updates to the test-app.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -919,6 +926,31 @@ private boolean validateUserId(String userId) {
...
     }
 
+    //======== Notification APIs ========//
+
+    public NotificationCenter getNotificationCenter() {
+        return notificationCenter;
+    }
+
+    /**
+     * Convenience method for adding TrackNotification Handlers
```
mikeng13: adding `DecisionNotification` handlers

core-api/src/main/java/com/optimizely/ab/notification/ActivateNotificationListener.java
```diff
@@ -22,10 +22,17 @@
 import com.optimizely.ab.event.LogEvent;
 
 import javax.annotation.Nonnull;
+import java.rmi.activation.ActivateFailedException;
```
mikeng13: Is this needed? I am not seeing it used below

core-api/src/main/java/com/optimizely/ab/notification/Notification.java
```diff
@@ -0,0 +1,26 @@
...
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+package com.optimizely.ab.notification;
+
+/**
+ * Interface used to identify supported annotations
```
mikeng13: Supported notifications?

core-api/src/main/java/com/optimizely/ab/notification/NotificationManager.java
```diff
@@ -0,0 +1,74 @@
...
+    private final Map<Integer, NotificationHandler<T>> handlers = new HashMap<>();
+    private final AtomicInteger counter;
+
+    public NotificationManager(AtomicInteger counter) {
+        this.counter = counter;
+    }
+
+    public int addHandler(NotificationHandler<T> newHandler) {
+
+        // Prevent registering a duplication listener.
```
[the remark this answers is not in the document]
loganlinn: Would it make sense to flip the map to be `Map<NotificationHandler, Int>`?

core-api/src/main/java/com/optimizely/ab/notification/NotificationManager.java
```diff
@@ -0,0 +1,74 @@
...
+ * class.
+ *
+ * The NotificationManager is responsible for storing a collection of NotificationHandlers and mapping
+ * them to a globally unique integer so that they can be removed on demand.
+ */
+public class NotificationManager<T> {
+
+    private static final Logger logger = LoggerFactory.getLogger(NotificationManager.class);
+
+    private final Map<Integer, NotificationHandler<T>> handlers = new HashMap<>();
```
loganlinn: Use `LinkedHashMap` so there’s consistent order in handler invocation?

core-api/src/main/java/com/optimizely/ab/notification/ActivateNotificationListener.java
```diff
@@ -51,6 +58,17 @@ public final void notify(Object... args) {
         onActivate(experiment, userId, attributes, variation, logEvent);
     }
 
+    @Override
+    public final void notify(ActivateNotification message) {
```
loganlinn: If it doesn’t break public API compatibility, it would nice to not be overloading `Object.notify` anymore...

core-api/src/main/java/com/optimizely/ab/notification/NotificationManager.java
```diff
@@ -0,0 +1,74 @@
...
+    private final AtomicInteger counter;
+
+    public NotificationManager(AtomicInteger counter) {
+        this.counter = counter;
```
loganlinn: Seems odd for this constructor to be `public`; could use a null check if it remains public.

core-api/src/main/java/com/optimizely/ab/notification/NotificationCenter.java
```diff
@@ -73,104 +72,63 @@ public Class getNotificationTypeClass() {
...
-
-        NotificationHolder(int id, NotificationListener notificationListener) {
-            notificationId = id;
-            this.notificationListener = notificationListener;
-        }
+    @Nonnull
+    @SuppressWarnings("unchecked")
+    public <T> NotificationManager<T> getNotificationManager(Class<? extends Notification> clazz) {
+        NotificationManager<T> newManager = new NotificationManager<>(counter);
+        NotificationManager<T> manager = (NotificationManager<T>) notifierMap.putIfAbsent(clazz, newManager);
```
loganlinn: Use `computeIfAbsent()` to avoid creating throw-away instance when manager exists? Alternatively, the set of notification types is (all likely will remain) trivially small, so could just pre-populate the map.

core-api/src/main/java/com/optimizely/ab/notification/NotificationCenter.java
```diff
@@ -181,75 +139,49 @@ public void onTrack(@Nonnull String eventKey, @Nonnull String userId, @Nonnull M
...
     }
 
     /**
      * Clear out all the notification listeners.
      */
     public void clearAllNotificationListeners() {
-        decisionListenerHolder.clear();
-        for (NotificationType type : NotificationType.values()) {
-            clearNotificationListeners(type);
+        for (NotificationManager<?> manager : notifierMap.values()) {
```
loganlinn: Minor: just clearing the entire `notifierMap` would be more consistent with the current scheme of lazy population used with `putIfAbsent`

core-api/src/main/java/com/optimizely/ab/notification/NotificationCenter.java
```diff
@@ -258,35 +190,27 @@ public void clearAllNotificationListeners() {
...
-            try {
-                holder.notificationListener.notify(args);
-            } catch (Exception e) {
-                logger.error("Unexpected exception calling notification listener {}", holder.notificationId, e);
-            }
+    @SuppressWarnings("unchecked")
+    public void send(Notification notification) {
+        NotificationManager handler = notifierMap.get(notification.getClass());
+        if (handler == null) {
+            return;
```
loganlinn: This seems worth logging a warn or error

core-api/src/main/java/com/optimizely/ab/notification/NotificationHandler.java
```diff
@@ -0,0 +1,21 @@
...
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+package com.optimizely.ab.notification;
+
+public interface NotificationHandler<T> {
+    void notify(T message);
```
loganlinn: Left a comment elsewhere but I’ll leave one here too: not a great name for any Java method as every Object has `notify()` for very different purpose.

core-api/src/main/java/com/optimizely/ab/notification/ActivateNotificationListener.java
mikecdavis: Ah, good point I can change that. Suggestion? `handle`? since the new interface is `NotificationHandler<T>`

mikeng13:
Passing custom build: https://travis-ci.com/optimizely/fullstack-sdk-compatibility-suite/jobs/197018131

core-api/src/main/java/com/optimizely/ab/notification/NotificationManager.java
mikecdavis: Made` package private`

core-api/src/main/java/com/optimizely/ab/notification/NotificationCenter.java
mikecdavis: Update: We're now creating a unmodifiable map containing all possible NotificationManager types, so clearing the individual managers is required.

core-api/src/main/java/com/optimizely/ab/notification/NotificationCenter.java
mikecdavis: Replaced with an `OptimizelyRuntimeException` since "we" control the notifications and should be catching this in our CI

core-api/src/main/java/com/optimizely/ab/notification/NotificationHandler.java
mikecdavis: Still need to do this.... sticking with "handle" since no other option has been proposed :)

mikecdavis:
@loganlinn as always, thanks for the feedback:) 

I ended up defining an unmodifiable Map of allowable notification types to control access in lieu of the `Notification` interface which proved difficult to constrain. So instead of lazily instantiating the set of `NotificationManager`s we're defining those up front to gate registration and notification sending and to fail fast during development of new notifications.

mikeng13:
Passing build after new changes here: https://travis-ci.com/optimizely/fullstack-sdk-compatibility-suite/builds/110421443

PR #292, opened 2019-05-03
Update HttpProjectConfigManager to use Last-Modified time

aliabbasrizvi:
## Summary
We now store and use the `Last-Modified` time when requesting the datafile. In all requests `If-Modified-Since` header is set and `HttpProjectConfigManager` creates new project config if there is a HTTP 200 response.


core-httpclient-impl/src/main/java/com/optimizely/ab/config/HttpProjectConfigManager.java
```diff
@@ -58,16 +54,58 @@ public URI getUri() {
         return uri;
     }
 
+    @VisibleForTesting
```
mikecdavis: Not needed, IMHO

core-httpclient-impl/src/test/java/com/optimizely/ab/config/HttpProjectConfigManagerTest.java
```diff
@@ -128,53 +142,64 @@ public void testBuildDefer() throws Exception {
...
-        handler.handleResponse(getResponse);
+        projectConfigManager.getDatafileFromResponse(getResponse);
     }
 
-    @Test
     public void testInvalidPayload() throws Exception {
         reset(mockHttpClient);
-        when(mockHttpClient.execute(any(HttpGet.class), any(ResponseHandler.class)))
-            .thenReturn("I am an invalid response!");
+        CloseableHttpResponse invalidPayloadResponse = mock(CloseableHttpResponse.class);
```
mikecdavis: Can we use the `BasicHttpResponse` here?
```
new BasicHttpResponse(new ProtocolVersion("TEST", 0, 0), 200, "I am an invalid response!");
```

aliabbasrizvi: `BasicHttpResponse` implements `HttpResponse` and not `CloseableHttpResponse` and so we can't use that here.

PR #295, opened 2019-05-06
Catch errors thrown while sending notifications.

mikecdavis:
## Summary
- Catch all exceptions thrown from NotificationHandlers.

This fixes a regression in #288 refactor when a `NotificationHandler` throws an error.


aliabbasrizvi:
Confirmed that unit tests and compatibility suite tests are passing.

PR #296, opened 2019-05-06
Making Optimizely Autocloseable

aliabbasrizvi:
Making Optimizely `Autocloseable` and calling close on `EventHandler` and `ProjectConfigManager` if they are `Closeable` themselves.

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/Optimizely.java
+++ b/core-api/src/main/java/com/optimizely/ab/Optimizely.java
@@ -122,17 +122,24 @@ public class Optimizely implements AutoCloseable {
    * are Closeable {@link Closeable} and calls close on them.
    *
    * <b>NOTE:</b> There is a chance that this could be long running if the implementations of close are long running.
-   *
-   * @throws IOException
    */
   @Override
-    public void close() throws IOException {
+    public void close() {
         if (eventHandler instanceof Closeable) {
-            ((Closeable) eventHandler).close();
+            try {
+                ((Closeable) eventHandler).close();
+            } catch (Exception e) {
+                logger.warn("Unexpected exception on trying to close event handler.");
+            }
+
         }
 
         if (projectConfigManager instanceof Closeable) {
-            ((Closeable) projectConfigManager).close();
+            try {
+                ((Closeable) projectConfigManager).close();
+            } catch (Exception e) {
+                logger.warn("Unexpected exception on trying to close config manager.");
+            }
         }
     }
 
--- a/core-api/src/main/java/com/optimizely/ab/Optimizely.java
+++ b/core-api/src/main/java/com/optimizely/ab/Optimizely.java
@@ -38,7 +38,6 @@ import javax.annotation.Nonnull;
 import javax.annotation.Nullable;
 import javax.annotation.concurrent.ThreadSafe;
 import java.io.Closeable;
-import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -115,6 +117,17 @@ public boolean isValid() {
         return getProjectConfig() != null;
     }
 
+    @Override
+    public void close() throws IOException {
```
aliabbasrizvi: Would love guidance on what sort of tests make sense here.

mikecdavis: * Test that close gets called for these implementations. 
* Test that exceptions thrown during close are handled appropriately (probably just a warn)
* There's also the case that they might block indefinitely trying to close, but maybe we need to leave that up to the implementations themselves to prevent. So a comment would suffice.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -115,6 +117,17 @@ public boolean isValid() {
...
+        if (this.eventHandler instanceof Closeable) {
```
mikecdavis: Nit, but I tend to only prefix with `this.` when it's necessary. Like resolving variable name conflicts.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
aliabbasrizvi: Thanks. Will respond accordingly.

core-api/src/test/java/com/optimizely/ab/OptimizelyTest.java
```diff
@@ -148,6 +144,44 @@ public OptimizelyTest(String validDatafile, String noAudienceDatafile) throws Co
...
+            .withConfigManager(mockCloseableProjectConfigManager)
+            .build();
+
+        optimizely.close();
+
+        verify((Closeable) mockCloseableEventHandler).close();
+        verify((Closeable) mockCloseableProjectConfigManager).close();
+    }
+
+    @Test(expected = IOException.class)
```
mikecdavis: We don't want to throw an exception. I think we should be catching all exceptions.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -115,6 +117,32 @@ public boolean isValid() {
...
+   * Checks if eventHandler {@link EventHandler} and projectConfigManager {@link ProjectConfigManager}
+   * are Closeable {@link Closeable} and calls close on them.
+   *
+   * <b>NOTE:</b> There is a chance that this could be long running if the implementations of close are long running.
+   */
+  @Override
+    public void close() {
+        if (eventHandler instanceof Closeable) {
+            try {
+                ((Closeable) eventHandler).close();
```
mikecdavis: Maybe best to create a tryClose(Object obj) method.
```
private void tryClose(Object obj) {
    if (!(obj instanceof Closeable)) {
        return;
    }
    
    try {
        ((Closeable) obj).close();
    } catch (Exception e) {
        logger.warn("Unexpected exception on trying to close {}.", obj);
    }
}
```

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -115,6 +117,32 @@ public boolean isValid() {
         return getProjectConfig() != null;
     }
 
+  /**
+   * Checks if eventHandler {@link EventHandler} and projectConfigManager {@link ProjectConfigManager}
+   * are Closeable {@link Closeable} and calls close on them.
+   *
+   * <b>NOTE:</b> There is a chance that this could be long running if the implementations of close are long running.
+   */
+  @Override
```
mikecdavis: Nit, spacing is not consistent.

PR #297, opened 2019-05-08
Fix: Deserializers to parse single leaf node audience condition

mnoman09:
## Summary
-  AudienceJacksonDeserializer was throwing exception on parsing single leaf node audience condition due to which datafile was not getting parsed.

## NOTE:
This is to support java sdk 2.1.x branch with latest datafile

## Ticket
- [OASIS-4568](https://optimizely.atlassian.net/browse/OASIS-4568)

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/config/parser/AudienceGsonDeserializer.java
+++ b/core-api/src/main/java/com/optimizely/ab/config/parser/AudienceGsonDeserializer.java
@@ -55,7 +55,7 @@ public class AudienceGsonDeserializer implements JsonDeserializer<Audience> {
         JsonElement conditionsElement = parser.parse(jsonObject.get("conditions").getAsString());
         Object rawObject = gson.fromJson(conditionsElement, Object.class);
         Condition conditions;
-        if (rawObject instanceof ArrayList) {
+        if (rawObject instanceof List) {
             conditions = parseConditions((List<Object>) rawObject);
         } else {
             conditions = parseConditions(rawObject);
@@ -95,7 +95,7 @@ public class AudienceGsonDeserializer implements JsonDeserializer<Audience> {
     private Condition parseConditions(Object rawObject) {
         List<Condition> conditions = new ArrayList<Condition>();
 
-        Map<String, String> conditionMap = (LinkedTreeMap<String, String>) rawObject;
+        Map<String, String> conditionMap = (Map<String, String>) rawObject;
         conditions.add(new UserAttribute(conditionMap.get("name"), conditionMap.get("type"),
                 conditionMap.get("value")));
 
--- a/core-api/src/main/java/com/optimizely/ab/config/parser/AudienceGsonDeserializer.java
+++ b/core-api/src/main/java/com/optimizely/ab/config/parser/AudienceGsonDeserializer.java
@@ -1,6 +1,6 @@
 /**
  *
- *    Copyright 2016-2017, Optimizely and contributors
+ *    Copyright 2016-2017, 2019, Optimizely and contributors
  *
  *    Licensed under the Apache License, Version 2.0 (the "License");
  *    you may not use this file except in compliance with the License.
@@ -36,7 +36,9 @@ import com.optimizely.ab.config.audience.UserAttribute;
 import java.lang.reflect.Type;
 
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 public class AudienceGsonDeserializer implements JsonDeserializer<Audience> {
 
@@ -51,8 +53,13 @@ public class AudienceGsonDeserializer implements JsonDeserializer<Audience> {
         String name = jsonObject.get("name").getAsString();
 
         JsonElement conditionsElement = parser.parse(jsonObject.get("conditions").getAsString());
-        List<Object> rawObjectList = gson.fromJson(conditionsElement, List.class);
-        Condition conditions = parseConditions(rawObjectList);
+        Object rawObject = gson.fromJson(conditionsElement, Object.class);
+        Condition conditions;
+        if (rawObject instanceof ArrayList) {
+            conditions = parseConditions((List<Object>) rawObject);
+        } else {
+            conditions = parseConditions(rawObject);
+        }
 
         return new Audience(id, name, conditions);
     }
@@ -81,6 +88,18 @@ public class AudienceGsonDeserializer implements JsonDeserializer<Audience> {
         } else {
             condition = new NotCondition(conditions.get(0));
         }
+
+        return condition;
+    }
+
+    private Condition parseConditions(Object rawObject) {
+        List<Condition> conditions = new ArrayList<Condition>();
+
+        Map<String, String> conditionMap = (LinkedTreeMap<String, String>) rawObject;
+        conditions.add(new UserAttribute(conditionMap.get("name"), conditionMap.get("type"),
+                conditionMap.get("value")));
+
+        Condition condition = new OrCondition(conditions);
 
         return condition;
     }
--- a/core-api/src/main/java/com/optimizely/ab/config/parser/AudienceJacksonDeserializer.java
+++ b/core-api/src/main/java/com/optimizely/ab/config/parser/AudienceJacksonDeserializer.java
@@ -1,6 +1,6 @@
 /**
  *
- *    Copyright 2016-2017, Optimizely and contributors
+ *    Copyright 2016-2017, 2019, Optimizely and contributors
  *
  *    Licensed under the Apache License, Version 2.0 (the "License");
  *    you may not use this file except in compliance with the License.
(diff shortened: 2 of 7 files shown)


core-api/src/main/java/com/optimizely/ab/config/parser/AudienceJacksonDeserializer.java
```diff
@@ -76,5 +82,17 @@ private Condition parseConditions(List<Object> rawObjectList) {
 
         return condition;
     }
+
+    private Condition parseConditions(Object rawObject) {
+        List<Condition> conditions = new ArrayList<Condition>();
+
+        HashMap<String, String> conditionMap = (HashMap<String, String>)rawObject;
+        conditions.add(new UserAttribute(conditionMap.get("name"), conditionMap.get("type"),
```
thomaszurkan-optimizely: name and type are non null values.  you should probably test the get name or have a default value setup.

mnoman09: In case where name or type is null then what should be the default value? 
currently we are getting illegal argument exception if we send null value in name or type.

mnoman09: @mikeng13

core-api/src/main/java/com/optimizely/ab/config/parser/AudienceGsonDeserializer.java
```diff
@@ -51,8 +53,13 @@ public Audience deserialize(JsonElement json, Type typeOfT, JsonDeserializationC
         String name = jsonObject.get("name").getAsString();
 
         JsonElement conditionsElement = parser.parse(jsonObject.get("conditions").getAsString());
-        List<Object> rawObjectList = gson.fromJson(conditionsElement, List.class);
-        Condition conditions = parseConditions(rawObjectList);
+        Object rawObject = gson.fromJson(conditionsElement, Object.class);
+        Condition conditions;
+        if (rawObject instanceof ArrayList) {
```
loganlinn:
```suggestion
        if (rawObject instanceof List) {
```

core-api/src/main/java/com/optimizely/ab/config/parser/AudienceGsonDeserializer.java
```diff
@@ -84,4 +91,16 @@ private Condition parseConditions(List<Object> rawObjectList) {
 
         return condition;
     }
+
+    private Condition parseConditions(Object rawObject) {
+        List<Condition> conditions = new ArrayList<Condition>();
+
+        Map<String, String> conditionMap = (LinkedTreeMap<String, String>) rawObject;
```
loganlinn:
```suggestion
        Map<String, String> conditionMap = (Map<String, String>) rawObject;
```

core-api/src/main/java/com/optimizely/ab/config/parser/AudienceJacksonDeserializer.java
```diff
@@ -43,25 +43,30 @@ public Audience deserialize(JsonParser parser, DeserializationContext context) t
 
         String id = node.get("id").textValue();
         String name = node.get("name").textValue();
-        List<Object> rawObjectList = (List<Object>)mapper.readValue(node.get("conditions").textValue(), List.class);
-        Condition conditions = parseConditions(rawObjectList);
 
+        Object rawObject = mapper.readValue(node.get("conditions").textValue(), Object.class);
+        Condition conditions;
+        if (rawObject instanceof ArrayList) {
```
loganlinn:
```suggestion
        if (rawObject instanceof List) {
```

core-api/src/main/java/com/optimizely/ab/config/parser/AudienceJacksonDeserializer.java
```diff
@@ -43,25 +43,30 @@ public Audience deserialize(JsonParser parser, DeserializationContext context) t
...
 
         for (int i = 1; i < rawObjectList.size(); i++) {
             Object obj = rawObjectList.get(i);
             if (obj instanceof List) {
-                List<Object> objectList = (List<Object>)rawObjectList.get(i);
+                List<Object> objectList = (List<Object>) rawObjectList.get(i);
                 conditions.add(parseConditions(objectList));
             } else {
-                HashMap<String, String> conditionMap = (HashMap<String, String>)rawObjectList.get(i);
+                HashMap<String, String> conditionMap = (HashMap<String, String>) rawObjectList.get(i);
```
loganlinn:
```suggestion
                Map<String, String> conditionMap = (Map<String, String>) rawObjectList.get(i);
```

core-api/src/main/java/com/optimizely/ab/config/parser/AudienceJacksonDeserializer.java
```diff
@@ -76,5 +81,17 @@ private Condition parseConditions(List<Object> rawObjectList) {
 
         return condition;
     }
+
+    private Condition parseConditions(Object rawObject) {
+        List<Condition> conditions = new ArrayList<Condition>();
+
+        HashMap<String, String> conditionMap = (HashMap<String, String>) rawObject;
```
loganlinn:
```suggestion
        HashMap<String, String> conditionMap = (Map<String, String>) rawObject;
```

core-api/src/main/java/com/optimizely/ab/config/parser/JsonConfigParser.java
```diff
@@ -266,19 +266,43 @@ public ProjectConfig parseProjectConfig(@Nonnull String json) throws ConfigParse
...
-            JSONObject audienceObject = (JSONObject)obj;
+            JSONObject audienceObject = (JSONObject) obj;
             String id = audienceObject.getString("id");
             String key = audienceObject.getString("name");
             String conditionString = audienceObject.getString("conditions");
-
-            JSONArray conditionJson = new JSONArray(conditionString);
-            Condition conditions = parseConditions(conditionJson);
+            Condition conditions;
+            if (conditionString.startsWith("[")) {
```
loganlinn: it would be better to use `JSONTokener` to inspect the JSON string versus directly inspecting (side-note: `charAt(0) == '['` should be preferred over `startsWith("[")`). The tokenizer can be passed to `JSONArray`/`JSONObject` constructor rather than string (it does this under the hood anyway)

core-api/src/main/java/com/optimizely/ab/config/parser/JsonConfigParser.java
```diff
@@ -266,19 +266,43 @@ public ProjectConfig parseProjectConfig(@Nonnull String json) throws ConfigParse
...
+    private Condition parseConditions(JSONObject conditionJson) {
+        List<Condition> conditions = new ArrayList<Condition>();
+
+        JSONObject conditionMap = conditionJson;
+        String value = null;
+        if (conditionMap.has("value")) {
+            value = conditionMap.getString("value");
+        }
+        conditions.add(new UserAttribute(
+                (String) conditionMap.get("name"),
```
loganlinn: we use `conditionMap.getString(...)` few lines above, but `(String) conditionMap.get(...)` here. Are those intentionally different?

Side-note: majority of the time, `foo.toString()` is better than `(String) foo`

core-api/src/test/java/com/optimizely/ab/config/ValidProjectConfigV4.java
```diff
@@ -114,6 +114,17 @@
             ))
     );
 
+    private static final String     UNTYPED_SINGLE_CONDITION_LEAF_ID = "10413101795";
+    private static final String     UNTYPED_SINGLE_CONDITION_LEAF_KEY = "untyped_single_condition_leaf_root";
+    public  static final String     UNTYPED_SINGLE_CONDITION_LEAF_VALUE = "leaf_root";
+    private static final Audience   UNTYPED_SINGLE_CONDITION_LEAF = new Audience(
```
loganlinn: nit: aligning code like this doesn't follow any existing code styles

core-api/src/main/java/com/optimizely/ab/config/parser/AudienceJacksonDeserializer.java
thomaszurkan-optimizely: @mnoman09 @mikeng13 default can probably be empty string for now.  but, i'm confused.  were single leaf conditions allowed in the datafile after typed audiences or is this a fix for compat?

mnoman09: here is the [Link](https://github.com/optimizely/fullstack-sdk-compatibility-suite/blob/6317c6a6b99e4bbcff84924a2c0e4a65233ea9b1/features/support/datafiles/audience_targeting.json#L1597) of line and datafile of compat suite in which we have single condition leaf root.  Due to these type of audience we were unable to parse datafile and all other testcases were failing which were not even related to this audience.

mnoman09:
Closing this PR as this change is no longer required.

PR #299, opened 2019-05-13
Create generic method implementation.

mikecdavis:
## Summary
- Create a generic addNotificaitonHandler, instead of the implementation specific ones.

Reduce the number of signatures on Optimizely. Use an interface to identify supported notification
types. Note this breaks methods newly added to 3.1, but makes new additions easier going forward.

## Test plan
Unit and FSC


aliabbasrizvi:
As discussed IRL, it may be better to have APIs to add different notification handlers all of which call into `addNotificationHandler`.

PR #308, opened 2019-06-05
Refactor OptimizelyTest to have less reliance on Mockito.

mikecdavis:
## Summary
- Removes mockito from LogbackVerifier
- Adds EventHandlerRule for validating dispatched events.
- Removes dependency on mocked classes for the majority of tests.
- Some stylized refactoring to condense statements.

The overuse of Mockito has made the tests fairly brittle and also required a large amount of setup for each specific test. Moving to a more black box approach and verify the side-effects of the scenarios makes for much cleaner and more concise tests.

This is also a prerequisite for a larger change to the event processing in which we'll be batching requests to the log endpoint. If this refactor was not done, then we'd have to change all of the mock setups related to dispatching at that time which would erode confidence that we're maintaining parity when introducing that change.

### DIFF
--- a/core-api/src/test/java/com/optimizely/ab/OptimizelyTest.java
+++ b/core-api/src/test/java/com/optimizely/ab/OptimizelyTest.java
@@ -21,33 +21,27 @@ import com.optimizely.ab.bucketing.Bucketer;
 import com.optimizely.ab.bucketing.DecisionService;
 import com.optimizely.ab.bucketing.FeatureDecision;
 import com.optimizely.ab.config.*;
-import com.optimizely.ab.config.parser.ConfigParseException;
-import com.optimizely.ab.error.ErrorHandler;
 import com.optimizely.ab.error.NoOpErrorHandler;
 import com.optimizely.ab.error.RaiseExceptionErrorHandler;
 import com.optimizely.ab.event.EventHandler;
 import com.optimizely.ab.event.LogEvent;
 import com.optimizely.ab.event.internal.EventFactory;
-import com.optimizely.ab.event.internal.payload.EventBatch;
 import com.optimizely.ab.internal.LogbackVerifier;
 import com.optimizely.ab.internal.ControlAttribute;
 import com.optimizely.ab.notification.*;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
-import org.junit.runners.Parameterized.Parameters;
-import org.mockito.ArgumentCaptor;
 import org.mockito.Mock;
 import org.mockito.junit.MockitoJUnit;
 import org.mockito.junit.MockitoRule;
 
-import javax.annotation.Nonnull;
 import java.io.Closeable;
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
@@ -65,8 +59,6 @@ import static java.util.Arrays.asList;
 import static junit.framework.TestCase.assertTrue;
 import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.hasEntry;
-import static org.hamcrest.Matchers.hasKey;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
@@ -81,20 +73,23 @@ import static org.mockito.Mockito.*;
 @RunWith(Parameterized.class)
 public class OptimizelyTest {
 
-    @Parameters
+    @Parameterized.Parameters(name = "{index}: Version: {2}")
     public static Collection<Object[]> data() throws IOException {
         return Arrays.asList(new Object[][]{
             {
                 validConfigJsonV2(),
                 noAudienceProjectConfigJsonV2(),
+                2
             },
             {
                 validConfigJsonV3(),
                 noAudienceProjectConfigJsonV3(),  // FIX-ME this is not a valid v3 datafile
+                3
             },
             {
                 validConfigJsonV4(),
-                validConfigJsonV4()
+                validConfigJsonV4(),
+                4
             }
         });
     }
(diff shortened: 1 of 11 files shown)


core-api/src/test/java/com/optimizely/ab/OptimizelyTest.java
```diff
@@ -2146,174 +1253,91 @@ public void trackEventWithNullOrEmptyUserID() throws Exception {
...
         logbackVerifier.expectMessage(Level.ERROR, "Unexpected exception in event dispatcher");
-        optimizely.track(eventType.getKey(), testUserId);
     }
 
     /**
      * Verify that {@link Optimizely#track(String, String, Map)}
      * dispatches events even if the event links only to launched experiments
      */
     @Test
+//    @Ignore("Is this a bug??")
```
aliabbasrizvi: Not a bug. Launched state works like experiment is running, but impression event is not dispatched during activate.

core-api/src/test/java/com/optimizely/ab/OptimizelyTest.java
```diff
@@ -2146,174 +1253,91 @@ public void trackEventWithNullOrEmptyUserID() throws Exception {
...
+        Optimizely optimizely = optimizelyBuilder.withConfig(noAudienceProjectConfig).build();
+        optimizely.track(eventType.getKey(), genericUserId);
+        eventHandler.expectConversion(eventType.getKey(), genericUserId);
     }
 
     /**
      * Verify that {@link Optimizely#track(String, String, Map)}
      * dispatches log events when the tracked event links to both launched and running experiments.
+     *
+     * TODO Is this test needed? Feels unnecessary in a post EET world.
```
aliabbasrizvi: +1

core-api/src/test/java/com/optimizely/ab/OptimizelyTest.java
```diff
@@ -3437,105 +2361,53 @@ public void activateWithListener() throws Exception {
...
-            @Override
-            public void onActivate(@Nonnull Experiment experiment, @Nonnull String userId, @Nonnull Map<String, ?> attributes, @Nonnull Variation variation, @Nonnull LogEvent event) {
-                assertEquals(experiment.getKey(), activatedExperiment.getKey());
-                assertEquals(bucketedVariation.getKey(), variation.getKey());
-                assertEquals(userId, testUserId);
-                for (Map.Entry<String, ?> entry : attributes.entrySet()) {
-                    assertEquals(testUserAttributes.get(entry.getKey()), entry.getValue());
-                }
-
-                assertEquals(event.getRequestMethod(), RequestMethod.GET);
```
aliabbasrizvi: Was this a bug?

core-api/src/test/java/com/optimizely/ab/event/NoopEventHandlerTest.java
```diff
@@ -1,6 +1,6 @@
 /**
  *
- *    Copyright 2016-2017, Optimizely and contributors
+ *    Copyright 2016-2017, 2019 Optimizely and contributors
```
aliabbasrizvi: Updated even when no change?

core-api/src/test/java/com/optimizely/ab/OptimizelyTest.java
mikecdavis: The mocked `logEventToDispatch` was a GET, removing the mock and using the actual `EventFactory` meant this needed to change to POST.

core-api/src/test/java/com/optimizely/ab/OptimizelyTest.java
mikecdavis: Need to remove that. I was investigating the test setup and got it working.

core-api/src/test/java/com/optimizely/ab/OptimizelyTest.java
```diff
@@ -1077,59 +634,29 @@ public void activateWithTypedAttributes() throws Exception {
...
-            .withBucketing(mockBucketer)
-            .withEventBuilder(mockEventFactory)
-            .withConfig(validProjectConfig)
-            .withErrorHandler(new RaiseExceptionErrorHandler())
-            .build();
+        Map<String, String> testUserAttributes = new HashMap<>();
+        testUserAttributes.put("unknownAttribute", "dimValue");
 
-        Map<String, String> testUserAttributes = new HashMap<String, String>();
         if (datafileVersion >= 4) {
```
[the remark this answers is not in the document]
mikecdavis: Yeah it might not be necessary at all. More of the copy-paste that needs to be purged.

core-api/src/test/java/com/optimizely/ab/OptimizelyTest.java
```diff
@@ -1077,59 +634,29 @@ public void activateWithTypedAttributes() throws Exception {
      * In this case, the activate call should not remove the unknown attribute from the given map.
      */
     @Test
+    @Ignore("Kind of a silly test since the event factory strips these unknownAttributes out before sending.")
```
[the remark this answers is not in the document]
mikecdavis: The test was originally asserting that the `unknownattribute` were getting passed to the EventFactory:
```
// setup the attribute map captor (so we can verify its content)
ArgumentCaptor<Map> attributeCaptor = ArgumentCaptor.forClass(Map.class);

// verify that the event builder was called with the expected attributes
verify(mockEventFactory).createImpressionEvent(eq(validProjectConfig), eq(activatedExperiment),
    eq(bucketedVariation), eq(testUserId), attributeCaptor.capture());

Map<String, String> actualValue = attributeCaptor.getValue();
assertThat(actualValue, hasKey("unknownAttribute"));
```
Now the test is going to assert that it's stripped from the payload, so the semantics have changed a bit.

core-api/src/test/java/com/optimizely/ab/OptimizelyTest.java
```diff
@@ -914,38 +558,28 @@ public void activateWithExperimentKey() throws Exception {
...
         Variation actualVariation = optimizely.activate(unknownExperiment.getKey(), testUserId);
-
-        // verify that null is returned, as no project config was available
         assertNull(actualVariation);
     }
 
     /**
      * Verify that {@link Optimizely#activate(String, String)} handles the case where an unknown experiment
      * (i.e., not in the config) is passed through and a {@link RaiseExceptionErrorHandler} is provided.
+     * DUPE? I think this is testing RaiseExceptionErrorHandler.
```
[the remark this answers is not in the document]
mikecdavis: Since that's the case I'm going to move this to a unit test.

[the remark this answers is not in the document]
mikecdavis: RaiseExceptionErrorHandlerTest unit test, which doesn't exist atm.

PR #309, opened 2019-06-07
Fixing how project config is updated. Also, fixing issue in Optimizely.close()

aliabbasrizvi:
## Summary
This change fixes:
- `Optimizely.close()` which currently checks that `eventHandler` and `projectConfigManager` are instances of `Closeable`. Changing that to check if they are instances of `AutoCloseable`.
- `ProjectConfig` is refetched every time in `getVariation`. Using the passing in config itself.
- `ProjectConfig` is fetched multiple times when `getEnabledFeatures` is called. Updating `isFeatureEnabled`.

## Test plan
- Unit tests.

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/Optimizely.java
+++ b/core-api/src/main/java/com/optimizely/ab/Optimizely.java
@@ -120,15 +120,15 @@ public class Optimizely implements AutoCloseable {
     }
 
     /**
-     * Helper method which checks if Object is an instance of Closeable and calls close() on it.
+     * Helper method which checks if Object is an instance of AutoCloseable and calls close() on it.
      */
     private void tryClose(Object obj) {
-        if (!(obj instanceof Closeable)) {
+        if (!(obj instanceof AutoCloseable)) {
             return;
         }
 
         try {
-            ((Closeable) obj).close();
+            ((AutoCloseable) obj).close();
         } catch (Exception e) {
             logger.warn("Unexpected exception on trying to close {}.", obj);
         }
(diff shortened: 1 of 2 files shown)


aliabbasrizvi:
I realize I have to fix 2 tests in `OptimizelyTest.java`. Following up with that.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -773,20 +781,20 @@ public Variation getVariation(@Nonnull Experiment experiment,
...
-        return getVariation(experiment, userId, attributes, getProjectConfig());
+        return getVariation(getProjectConfig(), experiment, userId, attributes);
     }
 
     @Nullable
-    public Variation getVariation(@Nonnull Experiment experiment,
-                                  @Nonnull String userId,
-                                  @Nonnull Map<String, ?> attributes,
-                                  @Nonnull ProjectConfig projectConfig) throws UnknownExperimentException {
+    private Variation getVariation(@Nonnull ProjectConfig projectConfig,
```
aliabbasrizvi: Making this private as public methods are not supposed to take `projectConfig`

mikecdavis: Technically backwards incompatible, but I'm fine with it.

aliabbasrizvi: Backwards incompatible because it was public in an alpha release? I think we should be comfortable removing this right away.

PR #310, opened 2019-06-13
Implement event batching.

mikecdavis:
## Summary
- Introduces an EventProcessor interface.
- Introduces a BatchEventProcessor (maybe should name it BatchingEventProcessor)
- Adds NotificationManager<EventBatch> to EventFactory.

Buffering events within a queue before dispatching is an optimization that should prevent SDK implementations from exhausting resources while increasing throughput. This implementation relies on a `BlockingQueue` to buffer events received from one-to-many producers. A single consumer thread continuously polls from this queue to build a batch before emitting the batched `LogEvent`.

Note that the `Optimizely` class only knows about the `EventProcessor` and `EventHandler` so that it can gracefully close those implementations, so not strictly necessary. All communication between components is handled via Notifications using the `NotificationManager<T>`. The `EventFactory` emits an `EventBatch` to a `NotificationManager<EventBatch>` which is subscribed to by the `EventProcessor`. The `EventProcessor` emits `LogEvents` to a `NotificationManager<LogEvent>` which is subscribed to by the `EventHandler` for eventual dispatching.

## Why notifications all the way down?
Part of this PR is experimenting with the idea of leveraging the `NotificationManager<T>` as a common conduit for message passing between components and that the SDK can be comprised of these various conduits to provide the full feature set. The advantage to this approach is that it allows for easier non-blocking async implementations to be built (if we have an `AsyncNotificaitonManager<T>` or `NotificaitonManager<T>#sendAsync` for example). It also centralizes "observability" points within the SDK. If we instrument and trace  `NotificationManager<T>` it _might_ be able to provide most, if not all of the observability we require.

This implementation of batching events was rather simple when working with this paradigm since each component didn't have to get directly linked together which made testing simple and removed a lot of implementation options that _could_ have been made in a non message passing world.

## Trade-offs / alternatives:
- We could queue events into the `EventProcessor` via direct calls with `Optimizely#track` and `Optimizely#sendImpression`.
- We could have the EventFactory accept and `EventProcessor` and queue batches directly.
- We could have the `EventProcessor` accept a callback when batches are ready to be emitted.
- We could have the `EventProcessor` accept an EventHandler directly.
- The relationship between components might be a little opaque.
- Notifications can be brittle relationships and dependent on timing and ordering of components and subscriptions. Explicit constructor injection forces well-ordering during construction.
- For Event processing I think the Notifications concept still holds, but it might not match when making Decisions for example which isn't fire-and-forget, but requires a synchronous response.


mikecdavis:
Note that a major distinction between this PR and the documented approach to batching is _where_ the queue is positioned in the flow. 

In this PR the queuing happens after the `EventBatch` object then immediately queued. Truly "batched" payloads are constructed from merging these objects together. In the prescribed doc we'd introduce another model similar to the `EventFactory#createImpressionEvent` and `EventFactory#createConversionEvent` signatures, queue those models consume those for batching.

I found in implementing that approach suffered from two problems. 
1) backwards compatibility with the current notification schema.
2) resolving the grouping strategy of what can be batched together.

The latter methodology does have some good principles to follow, I will create a subsequent PR that attempts to resolve the two above issues so we can compare the approaches and the tradeoffs.

mikecdavis:
Closed in favor of https://github.com/optimizely/java-sdk/pull/312

PR #312, opened 2019-06-21
Implement Event Batching Take2

mikecdavis:
## Summary
- Introduces an EventProcessor interface.
- Introduces a BatchEventProcessor (maybe should name it BatchingEventProcessor)
- Introduces an OptimizelyRule to handle lifecycle management during unit testing.

Buffering events within a queue before dispatching is an optimization that should prevent SDK implementations from exhausting resources while increasing throughput. This implementation relies on a `BlockingQueue` to buffer events received from one-to-many producers. A single consumer thread continuously polls from this queue to build a batch before emitting the batched `LogEvent`.

Note that the `Optimizely` class only knows about the `EventProcessor` and `EventHandler` so that it can gracefully close those implementations, so not strictly necessary. The Optimizely class builds UserEvents via the `UserEventFactory` and passes those events to the `EventProcessor`. The `EventProcessor` emits `LogEvents` to a `NotificationManager<LogEvent>` which is subscribed to by the `EventHandler` for eventual dispatching.

## Trade-offs / alternatives:
- We could have the `EventProcessor` accept an EventHandler directly, not leverging the NotificationManager<LogEvent>
- We could go all in on NotificationCenter as a global registry as opposed to one off implementations.

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
+++ b/core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
@@ -117,6 +117,8 @@ public class BatchEventProcessor implements EventProcessor, AutoCloseable {
     }
 
     public void process(UserEvent userEvent) {
+        logger.debug("Received userEvent: {}", userEvent);
+
 
         if (executor.isShutdown()) {
             logger.warn("Executor shutdown, not accepting tasks.");
@@ -143,7 +145,7 @@ public class BatchEventProcessor implements EventProcessor, AutoCloseable {
 
                     Object item = eventQueue.poll(50, TimeUnit.MILLISECONDS);
                     if (item == null) {
-                        logger.info("Empty item, sleeping for 500ms.");
+                        logger.info("Empty item, sleeping for 50ms.");
                         Thread.sleep(50);
                         continue;
                     }
--- a/core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
+++ b/core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
@@ -16,6 +16,7 @@
  */
 package com.optimizely.ab.event;
 
+import com.optimizely.ab.config.ProjectConfig;
 import com.optimizely.ab.event.internal.EventFactory;
 import com.optimizely.ab.event.internal.UserEvent;
 import com.optimizely.ab.internal.PropertyUtils;
@@ -128,7 +129,7 @@ public class BatchEventProcessor implements EventProcessor, AutoCloseable {
 
                     Object item = eventQueue.poll(50, TimeUnit.MILLISECONDS);
                     if (item == null) {
-                        logger.info("Empty item, sleeping for 50ms.");
+                        logger.debug("Empty item, sleeping for 50ms.");
                         Thread.sleep(50);
                         continue;
                     }
@@ -151,14 +152,9 @@ public class BatchEventProcessor implements EventProcessor, AutoCloseable {
         }
 
         private void addToBatch(UserEvent userEvent) {
-            if (!currentBatch.isEmpty()) {
-                String currentRevision = currentBatch.peekLast().getUserContext().getProjectConfig().getRevision();
-
-                // Separate batches based on config revision.
-                if (!currentRevision.equals(userEvent.getUserContext().getProjectConfig().getRevision())) {
-                    flush();
-                    currentBatch = new LinkedList<>();
-                }
+            if (shouldSplit(userEvent)) {
+                flush();
+                currentBatch = new LinkedList<>();
             }
 
             // Reset the deadline if starting a new batch.
@@ -170,6 +166,27 @@ public class BatchEventProcessor implements EventProcessor, AutoCloseable {
             if (currentBatch.size() >= batchSize) {
                 flush();
             }
+        }
+
+        private boolean shouldSplit(UserEvent userEvent) {
+            if (currentBatch.isEmpty()) {
+                return false;
+            }
+
+            ProjectConfig currentConfig = currentBatch.peekLast().getUserContext().getProjectConfig();
+            ProjectConfig newConfig = userEvent.getUserContext().getProjectConfig();
+
+            // Revisions should match
+            if (!currentConfig.getProjectId().equals(newConfig.getProjectId())) {
+                return true;
+            }
+
+            // Projects should match
+            if (!currentConfig.getRevision().equals(newConfig.getRevision())) {
+                return true;
+            }
+
+            return false;
         }
 
         private void flush() {
--- a/core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
+++ b/core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
@@ -199,27 +199,27 @@ public class BatchEventProcessor implements EventProcessor, AutoCloseable {
         private Long flushInterval = null;
         private ExecutorService executor = null;
 
-        public Builder setEventHandler(EventHandler eventHandler) {
+        public Builder withEventHandler(EventHandler eventHandler) {
             this.eventHandler = eventHandler;
             return this;
         }
 
-        public Builder setEventQueue(BlockingQueue<Object> eventQueue) {
+        public Builder withEventQueue(BlockingQueue<Object> eventQueue) {
             this.eventQueue = eventQueue;
             return this;
         }
 
-        public Builder setBatchSize(Integer batchSize) {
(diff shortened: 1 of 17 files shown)


core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
```diff
@@ -0,0 +1,204 @@
...
+        EventConsumer runnable = new EventConsumer();
+        future = executor.submit(runnable);
+    }
+
+    @Override
+    public void close() throws Exception {
+        logger.info("Start close");
+        eventQueue.put(SHUTDOWN_SIGNAL);
+        try {
+            future.get(5, TimeUnit.SECONDS);
```
mjc1283: Will batched events get lost when this is called? Should we attempt to send batched events at this point?

mikecdavis: This is a blocking call waiting for the task to complete, i.e. process all pending events. The task will gracefully stop once it receives the `SHUTDOWN_SIGNAL` message, so in theory there will be no data loss. 

That being said we can't block indefinitely so the timeout is there as a maximum wait time to flush on close, at which point events will get dropped. I should make that configurable since 5 seconds might be too tight if the buffer has a lot of pending events.

mikecdavis: Added https://optimizely.atlassian.net/browse/OASIS-4975 to track.

mjc1283: So we won't flush immediately, but instead just allow any scheduled flushes to occur up until the timeout? In JS, we flush immediately on close. Is there a reason not to flush immediately?

mikecdavis: The `SHUTDOWN_SIGNAL` will trigger a flush once consumed by the batching task. So this will be "immediate" if the queue is empty or it will be delayed until the consumer finishes draining the queue. I believe this is the proper behavior in a multi-threaded environment.

core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        private void addToBatch(UserEvent userEvent) {
+            if (!currentBatch.isEmpty()) {
+                String currentRevision = currentBatch.peekLast().getUserContext().getProjectConfig().getRevision();
+
+                // Separate batches based on config revision.
+                if (!currentRevision.equals(userEvent.getUserContext().getProjectConfig().getRevision())) {
+                    flush();
+                    currentBatch = new LinkedList<>();
```
aliabbasrizvi: Shouldn't there be a `currentBatch.add(userEvent)` here or better that we not return?

core-api/src/test/java/com/optimizely/ab/event/BatchEventProcessorTest.java
```diff
@@ -0,0 +1,125 @@
...
+    @Test
+    public void testDrainOnClose() throws Exception {
+        UserEvent userEvent = buildConversionEvent(EVENT_NAME);
+        eventProcessor.process(userEvent);
+        eventProcessor.close();
+
+        assertEquals(0, eventQueue.size());
+        eventHandlerRule.expectConversion(EVENT_NAME, USER_ID);
+    }
+
```
aliabbasrizvi: Can we add to test flush on differing revision number on new userEvent?

core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
mikecdavis: Good point, and lacks coverage. Will add a test!

core-api/src/main/java/com/optimizely/ab/event/ForwardingEventProcessor.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+public class ForwardingEventProcessor implements EventProcessor {
+
+    private static final Logger logger = LoggerFactory.getLogger(ForwardingEventProcessor.class);
+
+    private final NotificationManager<LogEvent> notificationManager = new NotificationManager<>();
+
+    @Override
+    public void process(UserEvent userEvent) {
+        logger.debug("Not processing event");
```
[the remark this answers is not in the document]
mikecdavis: It made sense at the time, but I'll remove.

core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
mikecdavis: Done

core-api/src/test/java/com/optimizely/ab/event/BatchEventProcessorTest.java
mikecdavis: Done

mikeng13:
Big question, but does Optimizely need to know about the event handler? Can we consider the event dispatching itself part of "processing"? If so, I'd say it's the top-level event processor's job to handle the lifecycle of the events, including dispatching them. That way could avoid having to introduce a brand new concept such as the `NotificationManager`.

aliabbasrizvi:
> Big question, but does Optimizely need to know about the event handler

I think given the current state of the world, yes because we allow customers to implement it.

mikecdavis:
@mikeng13 It's a valid question and something I added in the PR description. I'm fine not using the `NotificationManager` as the conduit between components, it was kind of an experiment in leveraging these constructs more internally. I'll have to keep the `EventHandler` as part of the `Optimizely.Builder` for backwards compatibility.

mikeng13:
> > Big question, but does Optimizely need to know about the event handler
> 
> I think given the current state of the world, yes because we allow customers to implement it.

I think for now, yes, because of backwards compat. But eventually I think we should decouple it from the client and make part of "event processing"


> @mikeng13 It's a valid question and something I added in the PR description. I'm fine not using the `NotificationManager` as the conduit between components, it was kind of an experiment in leveraging these constructs more internally. I'll have to keep the `EventHandler` as part of the `Optimizely.Builder` for backwards compatibility.

Yeah makes sense. Just curious what the result/learnings of the experiment is/are and whether it still makes sense to keep these new constructs (even if just internal) as we might see them as a paradigm to work towards to.

core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
```diff
@@ -0,0 +1,232 @@
...
+            try {
+                while (true) {
+                    if (System.currentTimeMillis() > deadline) {
+                        logger.debug("Deadline exceeded flushing current batch.");
+                        flush();
+                    }
+
+                    Object item = eventQueue.poll(50, TimeUnit.MILLISECONDS);
+                    if (item == null) {
+                        logger.info("Empty item, sleeping for 50ms.");
```
aliabbasrizvi: This might be very noisy at info level?

core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
```diff
@@ -0,0 +1,232 @@
...
+                logger.error("Uncaught exception processing buffer.", e);
+            } finally {
+                logger.info("Exiting processing loop. Attempting to flush pending events.");
+                flush();
+            }
+        }
+
+        private void addToBatch(UserEvent userEvent) {
+            if (!currentBatch.isEmpty()) {
+                String currentRevision = currentBatch.peekLast().getUserContext().getProjectConfig().getRevision();
```
aliabbasrizvi: One observation regarding this. The current set up allows me to create 1 instance of `BatchEventProcessor` and use it in multiple Optimizely instances each of which may be tied to different projects. It is certainly possible that a customer does such a thing.

In that case, I was thinking if:
- we want to somehow disallow that? or
- flush based on both revision and project ID.

core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
```diff
@@ -172,6 +168,27 @@ private void addToBatch(UserEvent userEvent) {
...
 
+        private boolean shouldSplit(UserEvent userEvent) {
+            if (currentBatch.isEmpty()) {
+                return false;
+            }
+
+            ProjectConfig currentConfig = currentBatch.peekLast().getUserContext().getProjectConfig();
+            ProjectConfig newConfig = userEvent.getUserContext().getProjectConfig();
+
+            // Revisions should match
```
aliabbasrizvi: nit. These comments are reversed.

PR #315, opened 2019-07-01
test(parser): Default parser

msohailhussain:
## Summary
Modified travis file to test all unit tests with default config parsers and default config parser is now set using ENV. 
All Parsers are defined in matrix to run all unit tests with each parser.

## Test plan
Unit tests should be pased
## Issues
- "THING-1234" or "Fixes #123"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,10 @@ install: true
 addons:
   srcclr: true
 env:
-  - DEFAULT_PARSER="com.google.gson.Gson"
-  - DEFAULT_PARSER="com.fasterxml.jackson.databind.ObjectMapper"
-  - DEFAULT_PARSER="org.json.JSONObject"
-  - DEFAULT_PARSER="org.json.simple.JSONObject"
+  - DEFAULT_PARSER=com.google.gson.Gson
+  - DEFAULT_PARSER=com.fasterxml.jackson.databind.ObjectMapper
+  - DEFAULT_PARSER=org.json.JSONObject
+  - DEFAULT_PARSER=org.json.simple.JSONObject
 script:
     - "./gradlew clean"
     - "./gradlew exhaustiveTest"
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,11 @@ jdk:
 install: true
 addons:
   srcclr: true
+env:
+  - DEFAULT_PARSER="com.google.gson.Gson"
+  - DEFAULT_PARSER="com.fasterxml.jackson.databind.ObjectMapper"
+  - DEFAULT_PARSER="org.json.JSONObject"
+  - DEFAULT_PARSER="org.json.simple.JSONObject"
 script:
     - "./gradlew clean"
     - "./gradlew exhaustiveTest"
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,10 @@ install: true
 addons:
   srcclr: true
 env:
-  - optimizely_default_parser=com.google.gson.Gson
-  - optimizely_default_parser=com.fasterxml.jackson.databind.ObjectMapper
-  - optimizely_default_parser=org.json.JSONObject
-  - optimizely_default_parser=org.json.simple.JSONObject
+  - optimizely_default_parser=GSON_CONFIG_PARSER
+  - optimizely_default_parser=JACKSON_CONFIG_PARSER
+  - optimizely_default_parser=JSON_CONFIG_PARSER
+  - optimizely_default_parser=JSON_SIMPLE_CONFIG_PARSER
 script:
     - "./gradlew clean"
     - "./gradlew exhaustiveTest"
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,10 @@ install: true
 addons:
   srcclr: true
 env:
-  - DEFAULT_PARSER=com.google.gson.Gson
-  - DEFAULT_PARSER=com.fasterxml.jackson.databind.ObjectMapper
-  - DEFAULT_PARSER=org.json.JSONObject
-  - DEFAULT_PARSER=org.json.simple.JSONObject
+  - optimizely_default_parser=com.google.gson.Gson
+  - optimizely_default_parser=com.fasterxml.jackson.databind.ObjectMapper
+  - optimizely_default_parser=org.json.JSONObject
+  - optimizely_default_parser=org.json.simple.JSONObject
 script:
     - "./gradlew clean"
     - "./gradlew exhaustiveTest"
--- a/core-api/src/main/java/com/optimizely/ab/config/parser/DefaultConfigParser.java
+++ b/core-api/src/main/java/com/optimizely/ab/config/parser/DefaultConfigParser.java
@@ -37,6 +37,23 @@ public final class DefaultConfigParser {
 
     //======== Helper methods ========//
 
+    public enum ConfigParsers {
+        GSON_CONFIG_PARSER("com.google.gson.Gson"),
+        JACKSON_CONFIG_PARSER("com.fasterxml.jackson.databind.ObjectMapper"),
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -7,6 +7,11 @@ jdk:
 install: true
 addons:
   srcclr: true
+env:
+  - optimizely_default_parser=com.google.gson.Gson
```
mikecdavis: We should use the enums here to make it clear there are a set number of supported parsers.

core-api/src/main/java/com/optimizely/ab/config/parser/DefaultConfigParser.java
```diff
@@ -37,6 +38,23 @@ public static ConfigParser getInstance() {
...
 
+    public enum ConfigParsers {
+        GSON_CONFIG_PARSER("com.google.gson.Gson"),
+        JACKSON_CONFIG_PARSER("com.fasterxml.jackson.databind.ObjectMapper"),
+        JSON_CONFIG_PARSER("org.json.JSONObject"),
+        JSON_SIMPLE_CONFIG_PARSER("org.json.simple.JSONObject");
+
+        private final String configParserValue;
+
+        ConfigParsers(String configParserValue) {
```
mikecdavis: Let's use a supplier in the constructor and change the enum name:
```
    public enum ConfigParserSupplier {
        GSON_CONFIG_PARSER("com.google.gson.Gson", GsonConfigParser::new),
        JACKSON_CONFIG_PARSER("com.fasterxml.jackson.databind.ObjectMapper", JacksonConfigParser::new),
        JSON_CONFIG_PARSER("org.json.JSONObject", JsonConfigParser::new),
        JSON_SIMPLE_CONFIG_PARSER("org.json.simple.JSONObject", JsonSimpleConfigParser::new);

        private final String className;
        private final Supplier<ConfigParser> supplier;

        ConfigParserSupplier(String className, Supplier<ConfigParser> supplier) {
            this.className = className;
            this.supplier = supplier;
        }

        ConfigParser get() {
            return supplier.get();
        }

        private boolean isPresent() {
            try {
                Class.forName(className);
                return true;
            } catch (ClassNotFoundException e) {
                return false;
            }
        }
    }
```

core-api/src/main/java/com/optimizely/ab/config/parser/DefaultConfigParser.java
```diff
@@ -47,7 +65,22 @@ public static ConfigParser getInstance() {
...
+                configParser = new GsonConfigParser();
+            } else if (ConfigParsers.JSON_SIMPLE_CONFIG_PARSER.toString().equals(configParserName)) {
+                configParser = new JsonSimpleConfigParser();
+            } else if (ConfigParsers.JSON_CONFIG_PARSER.toString().equals(configParserName)) {
+                configParser = new JsonConfigParser();
+            } else {
+                throw new MissingJsonParserException("unable to locate a JSON parser. "
+                    + "Please see <link> for more information");
+            }
+        } else if (isPresent("com.fasterxml.jackson.databind.ObjectMapper")) {
```
mikecdavis: Now that we have the enum as defined above this changes from `if...else if...`, to:
```
        for (ConfigParserSupplier supplier: ConfigParserSupplier.valuesI()) {
            if (!supplier.isPresent()) {
                continue;
            }

            ConfigParser configParser = supplier.get();
            logger.info("using json parser: {}", configParser.getClass().getSimpleName());
            return configParser;
        }

        throw new MissingJsonParserException("unable to locate a JSON parser. "
            + "Please see <link> for more information");
```

core-api/src/test/java/com/optimizely/ab/config/parser/DefaultConfigParserTest.java
```diff
@@ -16,15 +16,76 @@
...
-    public void createThrowException() throws Exception {
-        // FIXME - mdodsworth: hmmm, this isn't going to be the easiest thing to test
+    public void testPropertyDefaultParser() throws Exception {
+        String defaultParser = PropertyUtils.get("default_parser");
+        ConfigParser configParser = DefaultConfigParser.getInstance();
+        ProjectConfig actual = configParser.parseProjectConfig(validDatafile);
+        ProjectConfig expected = validProjectConfig;
+        verifyProjectConfig(actual, expected);
+        if(defaultParser != null) {
+            if (DefaultConfigParser.ConfigParsers.GSON_CONFIG_PARSER.toString().equals(defaultParser)) {
```
mikecdavis: with the enum as defined earlier we can use `ConfigParserSupplier.valueOf(defaultParser)`.

core-api/src/test/java/com/optimizely/ab/config/parser/DefaultConfigParserTest.java
```diff
@@ -16,15 +16,76 @@
...
+                Assert.assertThat(configParser, CoreMatchers.instanceOf(GsonConfigParser.class));
+            } else if (DefaultConfigParser.ConfigParsers.JACKSON_CONFIG_PARSER.toString().equals(defaultParser)) {
+                Assert.assertThat(configParser, CoreMatchers.instanceOf(JacksonConfigParser.class));
+            } else if (DefaultConfigParser.ConfigParsers.JSON_CONFIG_PARSER.toString().equals(defaultParser)) {
+                Assert.assertThat(configParser, CoreMatchers.instanceOf(JsonConfigParser.class));
+            } else if (DefaultConfigParser.ConfigParsers.JSON_SIMPLE_CONFIG_PARSER.toString().equals(defaultParser)) {
+                Assert.assertThat(configParser, CoreMatchers.instanceOf(JsonSimpleConfigParser.class));
+            }
+        }
```
mikecdavis: can we should have an `else` statement defaulting to the first enum that matches? not sure if this will be deterministic however, but worth a shot.

mnoman09: yes, we can do that first enum that matches will be GSON_CONFIG_PARSER

core-api/src/test/java/com/optimizely/ab/config/parser/DefaultConfigParserTest.java
```diff
@@ -16,15 +16,80 @@
...
+    public void testPropertyDefaultParser() throws Exception {
+        String defaultParser = PropertyUtils.get("default_parser");
+        ConfigParser configParser = DefaultConfigParser.getInstance();
+        ProjectConfig actual = configParser.parseProjectConfig(validDatafile);
+        ProjectConfig expected = validProjectConfig;
+        verifyProjectConfig(actual, expected);
+        if(defaultParser != null) {
+            DefaultConfigParser.ConfigParserSupplier defaultParserSupplier = DefaultConfigParser.ConfigParserSupplier.valueOf(defaultParser);
+
+            if (DefaultConfigParser.ConfigParserSupplier.GSON_CONFIG_PARSER.equals(defaultParserSupplier)) {
```
mikecdavis: Should be a `switch` statement.
```
Class expectedParser = GsonConfigParser.class;

if(defaultParser != null) {
    DefaultConfigParser.ConfigParserSupplier defaultParserSupplier = DefaultConfigParser.ConfigParserSupplier.valueOf(defaultParser);
    switch (defaultParserSupplier) {
        case GSON_CONFIG_PARSER:
            expectedParser = GsonConfigParser.class;
            break;
        case JACKSON_CONFIG_PARSER:
            expectedParser = JacksonConfigParser.class;
            break;
        case JSON_CONFIG_PARSER:
            expectedParser = JsonConfigParser.class;
            break;
        case JSON_SIMPLE_CONFIG_PARSER:
            expectedParser = JsonSimpleConfigParser.class;
            break;
        default:
            fail("Not a valid config parser");
    }
}

Assert.assertThat(configParser, CoreMatchers.instanceOf(expectedParser));
```

PR #318, opened 2019-07-11
 Add LogEvent notification.

mikecdavis:
## Summary
- Add explicit notification for LogEvent.
- Deprecate LogEvent in track and activate notifications.

With the introduction of the BatchEventProcessor we no longer have a one-to-one mapping of Optimizely API calls to LogEvents dispatched. This PR adds a notification every time a LogEvent is built and submitted to the EventHandler. This can be subscribed to by adding a `NotificationHandler<LogEvent>` directly to the `NotificationCenter` or via the `Optimizely#addLogEventNotificationHandler(NotificationHandler)` convenience method.

We're also deprecating the LogEvent as a member of the `TrackNotification` and `ActivateNotification` models.

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/Optimizely.java
+++ b/core-api/src/main/java/com/optimizely/ab/Optimizely.java
@@ -964,6 +964,13 @@ public class Optimizely implements AutoCloseable {
     }
 
     /**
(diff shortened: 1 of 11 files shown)


aliabbasrizvi:
This is great. @mjc1283 I think we should introduce this in JavaScript SDK as well.

PR #319, opened 2019-07-11
Add more control over BatchEventProcessor.

mikecdavis:
## Summary
- Add flush method to BatchEventProcessor.
- Allow BatchEventProcessor to be restarted.
- Add option to build BatchEventProcessor without starting the processing thread.


core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
```diff
@@ -142,6 +147,12 @@ public void run() {
                         break;
                     }
 
+                    if (item == FLUSH_SIGNAL) {
+                        logger.debug("Received flush signal.");
+                        flush();
+                        break;
```
mjc1283: Is it correct to break here? Then the event consumer will exit, and no more events will be consumed, right?

mikecdavis: Yikes. Yeah should be a `continue;` 🤦‍♂

PR #322, opened 2019-07-22
Incorporate batch processing into OptimizelyFactory.

mikecdavis:
## Summary
- Default to using `BatchEventProcessor` when using `OptimizelyFactory`.
- Lower the default batch size to reduce impact of an SDK upgrade.


mikecdavis:
Will update to 30 seconds and merge.

PR #327, opened 2019-07-30
Add clear deprecation path from factory builder method.

mikecdavis:
## Summary
- Make `Optimizely.Builder#withDatafile` public.
- Add docs articulating the migration path away from the current factory method.

This should get patched into version `3.2.1` and will allow the Android SDK to cleanly use non-deprecated methods.


mikecdavis:
there are already unit tests when `protected`

PR #331, opened 2019-08-12
Defend against invalid BatchEventProcessor configuration overrides.

mikecdavis:
## Summary
- Reverts to default config values when invalid for BatchEventProcessor.
- Adds expectedCall count to EventHandlerRule to simplify tests.

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
+++ b/core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
@@ -58,9 +58,9 @@ public class BatchEventProcessor implements EventProcessor, AutoCloseable {
     private final BlockingQueue<Object> eventQueue;
     private final EventHandler eventHandler;
 
-    private final int batchSize;
-    private final long flushInterval;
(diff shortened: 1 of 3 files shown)


core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
```diff
@@ -240,42 +230,64 @@ public static Builder builder() {
...
+        /**
+         * BatchSize is the maximum number of events contained within a single event batch.
+         */
         public Builder withBatchSize(Integer batchSize) {
             this.batchSize = batchSize;
             return this;
         }
 
+        /**
+         * FlushInterval is the maximum duration, in milliseconds, that an event will remain in flight before
```
aliabbasrizvi: nit. May be say event will remain in queue?

mikecdavis: I didn't want to confuse the pending batch with the `eventQueue`. There are no guarantees about how long an event might exist in the `eventQueue`.

PR #337, opened 2019-09-24
Refactor Decision service: (Audience Evaluator)

fayyazarshad:
## Summary
Added Audience Evaluator Service

## Test plan
Added test in ab.decision.audience.AudienceEvaluatorTest

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/decision/evaluator/DecisionAudienceEvaluator.java
+++ b/core-api/src/main/java/com/optimizely/ab/decision/evaluator/DecisionAudienceEvaluator.java
@@ -49,8 +49,8 @@ public class DecisionAudienceEvaluator implements AudienceEvaluator {
         if (experiment.getAudienceConditions() != null) {
             return evaluateAudienceConditions(experiment, userContext);
         } else {
-            Boolean resolveReturn = evaluateAudience(experiment, userContext);
-            return Boolean.TRUE.equals(resolveReturn);
+            Boolean isAudienceValid = evaluateAudience(experiment, userContext);
+            return Boolean.TRUE.equals(isAudienceValid);
         }
     }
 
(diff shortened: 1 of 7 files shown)


core-api/src/main/java/com/optimizely/ab/decision/evaluator/DecisionAudienceEvaluator.java
```diff
@@ -0,0 +1,108 @@
...
+     * @param userContext It have user id, attributes and a reference to the current {@link ProjectConfig}
+     * @return flag if user satisfies audience conditions for the experiment
+     */
+    @Override
+    public boolean evaluate(@Nonnull Experiment experiment,
+                            @Nonnull UserContext userContext) {
+        if (experiment.getAudienceConditions() != null) {
+            return evaluateAudienceConditions(experiment, userContext);
+        } else {
+            Boolean resolveReturn = evaluateAudience(experiment, userContext);
```
msohailhussain: Change the variable name.

core-api/src/main/java/com/optimizely/ab/decision/evaluator/DecisionAudienceEvaluator.java
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * Evaluate audience if experiment has no conditions
+     *
+     * @param experiment  The Experiment the user will be bucketed into.
+     * @param userContext It have user id, attributes and a reference to the current {@link ProjectConfig}
+     * @return flag if audience are valid for experiment
+     */
+    @Nullable
+    private static Boolean evaluateAudience(@Nonnull Experiment experiment,
```
msohailhussain: we were using annotation `visiblefortesting` just curious where is that?

fayyazarshad: As far as I have seen, we were not using visiblefortesting annotation on this method.

PR #343, opened 2019-10-18
(chore): refactor batch event processor to use blocking queue poll and take so as not to spin too much.

thomaszurkan-optimizely:
## Summary
- pass in current time - flush interval to polling mechanism.
- use take instead of poll if we have been cycling more than twice.
- reset deadline once we hit it so it doesn't repeat until next flush interval.

## Test plan
FSC test?

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
+++ b/core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
@@ -48,6 +48,7 @@ public class BatchEventProcessor implements EventProcessor, AutoCloseable {
     public static final String CONFIG_CLOSE_TIMEOUT  = "event.processor.close.timeout";
 
     public static final int DEFAULT_QUEUE_CAPACITY    = 1000;
+    public static final int DEFAULT_WAIT_COUNT        = 2;
     public static final int DEFAULT_BATCH_SIZE        = 10;
     public static final long DEFAULT_BATCH_INTERVAL   = TimeUnit.SECONDS.toMillis(30);
     public static final long DEFAULT_TIMEOUT_INTERVAL = TimeUnit.SECONDS.toMillis(5);
@@ -122,6 +123,10 @@ public class BatchEventProcessor implements EventProcessor, AutoCloseable {
         eventQueue.put(FLUSH_SIGNAL);
     }
 
+    private interface QueueService {
+        Object get() throws InterruptedException;
+    }
+
     public class EventConsumer implements Runnable {
         private LinkedList<UserEvent> currentBatch = new LinkedList<>();
         private long deadline = System.currentTimeMillis() + flushInterval;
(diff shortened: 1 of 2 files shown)


core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
```diff
@@ -122,26 +123,46 @@ public void flush() throws InterruptedException {
...
     public class EventConsumer implements Runnable {
         private LinkedList<UserEvent> currentBatch = new LinkedList<>();
         private long deadline = System.currentTimeMillis() + flushInterval;
 
         @Override
         public void run() {
             try {
+                int waitCount = 0;
+
+                QueueService polling = () -> eventQueue.poll(flushInterval, TimeUnit.MILLISECONDS);
```
mjc1283: When polling the queue, should we block until the next deadline, instead of blocking for `flushInterval`?

core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
```diff
@@ -122,26 +123,46 @@ public void flush() throws InterruptedException {
...
+
                 while (true) {
                     if (System.currentTimeMillis() > deadline) {
                         logger.debug("Deadline exceeded flushing current batch.");
                         flush();
+                        deadline = System.currentTimeMillis() + flushInterval;
                     }
 
-                    Object item = eventQueue.poll(50, TimeUnit.MILLISECONDS);
+                    Object item = using.get();
```
mikecdavis: I think the `QueueService` adds a bit too much cognitive load  then what we need in this already complex loop. Since we're basing our decision to use `take()` or `poll(...)` based on the number of times we're exited the poll with `null` items we can use a ternary operator inline:
```java
long timeout = deadline - System.currentTimeMillis();
Object item = emptyCount > MAX_EMPTY_COUNT ? eventQueue.take() : eventQueue.poll(timeout, TimeUnit.MILLISECONDS);
```

Note I'm suggesting renaming from "wait" to "empty" since I think better captures the intention of the counter. I also used the dynamic timeout as opposed to the fix `flushInterval` as we discussed earlier.

core-api/src/main/java/com/optimizely/ab/event/BatchEventProcessor.java
```diff
@@ -122,26 +123,44 @@ public void flush() throws InterruptedException {
...
     public class EventConsumer implements Runnable {
         private LinkedList<UserEvent> currentBatch = new LinkedList<>();
         private long deadline = System.currentTimeMillis() + flushInterval;
 
         @Override
         public void run() {
             try {
+                int waitCount = 0;
+
+                QueueService polling = () -> eventQueue.poll(System.currentTimeMillis() - flushInterval, TimeUnit.MILLISECONDS);
```
mikecdavis: this is not the same as `deadline - System.currentTimeMillis()`

PR #348, opened 2019-12-03
feat: getOptimizelyConfig API to get static experiments and features data

fayyazarshad:
## Summary
Added `OptimizelyConfigService` which generates `OptimizelyConfig` Object containing experiments and features data. This is the first step in the development of `getOptimizelyConfig` api. This service always generates a new `OptimzelyConfig` object for now. Next PR will contain caching implementation for `OptimizelyConfig` object.

## Test plan
Added Unit Tests


core-api/src/main/java/com/optimizely/ab/optimizelyconfig/OptimizelyConfigService.java
```diff
@@ -0,0 +1,191 @@
...
+        
+        List<String> featureKeys = projectConfig.getExperimentFeatureKeyMapping().get(experimentId);
+        String featureKey = featureKeys != null ? featureKeys.get(0) : null;
+
+        List<FeatureFlag> featureFlags = projectConfig.getFeatureFlags();
+        if (featureFlags != null) {
+            for (FeatureFlag featureFlag : featureFlags) {
+                featureIDVariablesMap.put(featureFlag.getKey(), featureFlag.getVariables());
+            }
+        }
```
jaeopt: We do not need to get variables for all featureFlags - just getVariables() for the target featureKey.

core-api/src/test/java/com/optimizely/ab/optimizelyconfig/OptimizelyConfigTest.java
```diff
@@ -0,0 +1,118 @@
...
+                Map<String, OptimizelyVariation> variationsMap = experiment.getVariationsMap();
+                variations.forEach(variation ->
+                    featureFlag.getVariables().forEach(variable -> {
+                        OptimizelyVariable optimizelyVariable = variationsMap.get(variation.getKey()).getVariablesMap().get(variable.getKey());
+                        assertEquals(variable.getId(), optimizelyVariable.getId());
+                        assertEquals(variable.getKey(), optimizelyVariable.getKey());
+                        assertEquals(variable.getType().getVariableType().toLowerCase(), optimizelyVariable.getType());
+                        if (!variation.getFeatureEnabled()) {
+                            assertEquals(variable.getDefaultValue(), optimizelyVariable.getValue());
+                        }
```
jaeopt: Can we add assert for "else" case which is more interesting here?

core-api/src/test/java/com/optimizely/ab/optimizelyconfig/OptimizelyConfigTest.java
```diff
@@ -0,0 +1,118 @@
...
+                    })
+                );
+            });
+        });
+    }
+
+    @Test
+    public void shouldReturnCorrectRevision() throws Exception {
+        String revision = optimizelyConfig.getRevision();
+        assertEquals(revision, revision);
```
jaeopt: Compare to self :)

core-api/src/test/java/com/optimizely/ab/OptimizelyTest.java
```diff
@@ -250,6 +250,13 @@ public void activateEndToEndWithTypedAudienceInt() throws Exception {
         eventHandler.expectImpression(activatedExperiment.getId(), actualVariation.getId(), testUserId, testUserAttributes);
     }
 
+    @Test
+    public void testOptimizelyConfig() throws Exception {
+        Optimizely optimizely = optimizelyBuilder.build();
+        OptimizelyConfig optimizelyConfig = optimizely.getOptimizelyConfig();
+        assertNotNull(optimizelyConfig);
+    }
+
```
jaeopt: Can we add a test case for returning null for config error just like other APIs?

core-api/src/test/java/com/optimizely/ab/optimizelyconfig/OptimizelyConfigTest.java
zashraf1985: Haha! @jaeopt this is still work in progress and has many stupid peices of code. please do not review it yet :)

core-api/src/main/java/com/optimizely/ab/optimizelyconfig/OptimizelyConfigService.java
```diff
@@ -0,0 +1,189 @@
...
+ * See the License for the specific language governing permissions and      *
+ * limitations under the License.                                           *
+ ***************************************************************************/
+package com.optimizely.ab.optimizelyconfig;
+
+import com.optimizely.ab.config.*;
+
+import java.util.*;
+
+public class OptimizelyConfigService {
```
jaeopt: This should not be public

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -454,6 +456,19 @@ public Boolean getFeatureVariableBoolean(@Nonnull String featureKey,
         );
     }
 
+    /**
+     * Get {@link OptimizelyConfig} containing experiments and features map
+     *
+     * @return {@link OptimizelyConfig}
+     */
+    public OptimizelyConfig getOptimizelyConfig() {
```
jaeopt: I see this new API is located in the middle of getVariable section. Can we add at the end of other APIs as a separate section?

core-api/src/main/java/com/optimizely/ab/optimizelyconfig/OptimizelyConfigService.java
```diff
@@ -0,0 +1,189 @@
...
+                featureKeyToVariablesMap.put(featureFlag.getKey(), featureFlag.getVariables())
+            );
+        }
+        return featureKeyToVariablesMap;
+    }
+
+    private String getExperimentFeatureKey(String experimentId) {        
+        List<String> featureKeys = projectConfig.getExperimentFeatureKeyMapping().get(experimentId);
+        String featureKey = featureKeys != null ? featureKeys.get(0) : null;
+        return featureKey;
```
mikecdavis: Nit:
```java
return featureKeys != null ? featureKeys.get(0) : null;
```

core-api/src/main/java/com/optimizely/ab/optimizelyconfig/OptimizelyConfigService.java
```diff
@@ -0,0 +1,189 @@
...
+                OptimizelyExperiment optimizelyExperiment = new OptimizelyExperiment(
+                    experiment.getId(),
+                    experiment.getKey(),
+                    getVariationsMap(experiment.getVariations(), experiment.getId())
+                );
+                experimentsMap.put(experiment.getKey(), optimizelyExperiment);
+            });
+        }
+        return experimentsMap;
+    }
```
mikecdavis: I think these methods can benefit from Java 8 streams:
```java
private Map<String, OptimizelyExperiment> getExperimentsMap() {
    List<Experiment> experiments = projectConfig.getExperiments();
    if (experiments == null) {
        return Collections.emptyMap();
    }

    return experiments.stream()
        .collect(Collectors.toMap(Experiment::getKey, experiment -> new OptimizelyExperiment(
            experiment.getId(),
            experiment.getKey(),
            getVariationsMap(experiment.getVariations(), experiment.getId())
        )));
}
```

core-api/src/test/java/com/optimizely/ab/optimizelyconfig/OptimizelyConfigTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+    private OptimizelyConfig optimizelyConfig;
+    private Optimizely optimizely;
+    private ProjectConfig projectConfig;
+
+    @Before
+    public void initialize()throws Exception {
+        projectConfig = new DatafileProjectConfig.Builder().withDatafile(validConfigJsonV4()).build();
+        optimizely = Optimizely.builder().withDatafile(validConfigJsonV4()).build();
+        optimizelyConfig = optimizely.getOptimizelyConfig();
```
mikecdavis: As a unit test, we shouldn't have a dependency on `Optimizely`. (It's also not needed as a class member anyway) 
We should be testing the service directly.
```java
projectConfig = new DatafileProjectConfig.Builder().withDatafile(validConfigJsonV4()).build();
optimizelyConfig = new OptimizelyConfigService(projectConfig).getOptimizelyConfig();
```

core-api/src/test/java/com/optimizely/ab/optimizelyconfig/OptimizelyConfigTest.java
```diff
@@ -0,0 +1,115 @@
...
+        assertEquals(optimizelyFeatureMap.size(), 7);
+
+        projectConfig.getFeatureFlags().forEach(featureFlag -> {
+            OptimizelyFeature optimizelyFeature = optimizelyFeatureMap.get(featureFlag.getKey());
+            assertNotNull(optimizelyFeature);
+
+            Map<String, OptimizelyVariable> variablesMap = optimizelyFeatureMap.get(featureFlag.getKey()).getVariablesMap();
+            featureFlag.getVariables().forEach(variable -> {
+                OptimizelyVariable optimizelyVariable = variablesMap.get(variable.getKey());
+                assertNotNull(optimizelyFeature);
```
mikecdavis: Should be `assertNotNull(optimizelyVariable);`?

core-api/src/test/java/com/optimizely/ab/optimizelyconfig/OptimizelyConfigTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+    @Test
+    public void shouldReturnAllExperimentsExceptRollouts() throws Exception {
+        Map<String, OptimizelyExperiment> optimizelyExperimentMap = optimizelyConfig.getExperimentsMap();
+        assertEquals(optimizelyExperimentMap.size(), 13);
+
+        List<Experiment> experiments = getAllExperimentsFromDatafile();
+        experiments.forEach(experiment -> {
+            OptimizelyExperiment optimizelyExperiment = optimizelyExperimentMap.get(experiment.getKey());
+            assertNotNull(optimizelyExperiment);
```
mikecdavis: This isn't asserting much, a lot can be going wrong and this won't catch it. We should be asserting on the returned value.

core-api/src/test/java/com/optimizely/ab/optimizelyconfig/OptimizelyConfigTest.java
```diff
@@ -0,0 +1,115 @@
...
+                List<Variation> variations = datafileExperimentsMap.get(experimentKey).getVariations();
+                Map<String, OptimizelyVariation> variationsMap = experiment.getVariationsMap();
+                variations.forEach(variation ->
+                    featureFlag.getVariables().forEach(variable -> {
+                        OptimizelyVariable optimizelyVariable = variationsMap.get(variation.getKey()).getVariablesMap().get(variable.getKey());
+                        assertEquals(variable.getId(), optimizelyVariable.getId());
+                        assertEquals(variable.getKey(), optimizelyVariable.getKey());
+                        assertEquals(variable.getType().getVariableType().toLowerCase(), optimizelyVariable.getType());
+                        if (!variation.getFeatureEnabled()) {
+                            assertEquals(variable.getDefaultValue(), optimizelyVariable.getValue());
```
mikecdavis: Nit, but this is a bear to read. I don't think the `.forEach(...)` is buying us much except making is more difficult to follow.

zashraf1985:
@mikecdavis ... Thanks for your feedback... You are right about unit tests not looking like the way they should in java. I should have been more watchful about it. I have been juggling between reviewing and implementing optimizely config on multiple languages so may be i mostly leaned towards porting some or most of the stuff instead of totally rethinking how it should be implemented somewhat differently in different languages. This is valuable feedback for me both in context of the current feature and also for the future work.

zashraf1985:
@jaeopt Thanks for the feedback.. We are sort of redoing the unit tests today so we will definitely add more validations as per your suggestions

core-api/src/main/java/com/optimizely/ab/optimizelyconfig/OptimizelyConfigService.java
```diff
@@ -15,89 +15,100 @@
...
     }
 
+    /**
+     * creates new instance if the revision is updated
+     *
+     * @param projectConfig project config provided
+     * @return singleton object of {@link OptimizelyConfigService}
+     */
+    public static OptimizelyConfigService getInstance(ProjectConfig projectConfig) {
+        if (serviceInstance == null || !projectConfig.getRevision().equals(serviceInstance.getOptimizelyConfig().getRevision())) {
```
jaeopt: We should check sdkKey value as well here since different projectConfig can have the same revision numbers accidentally.

jaeopt: More important issue here is that we need to keep a cache per sdkKey, isn't it? 
We need a general cache management solution to support multiple OptConfig instances.

core-api/src/test/java/com/optimizely/ab/optimizelyconfig/OptimizelyConfigServiceTest.java
```diff
@@ -0,0 +1,232 @@
...
+            assertEquals(optimizelyVariation.getKey(), variation.getKey());
+            Map<String, OptimizelyVariable> optimizelyVariableMap =
+                optimizelyConfigService.getMergedVariablesMap(variation, experiment.getId());
+            optimizelyVariableMap.forEach((variableKey, variable) -> {
+                OptimizelyVariable optimizelyVariable = optimizelyVariableMap.get(variableKey);
+                assertEquals(variable.getValue(), optimizelyVariable.getValue());
+                assertEquals(variable.getType(), optimizelyVariable.getType());
+                assertEquals(variable.getId(), optimizelyVariable.getId());
+                assertEquals(variable.getKey(), optimizelyVariable.getKey());
+            });
```
jaeopt: This is weird - looks like validating against to itself

core-api/src/main/java/com/optimizely/ab/optimizelyconfig/OptimizelyConfigService.java
```diff
@@ -0,0 +1,185 @@
...
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+public class OptimizelyConfigService {
+
+    private ProjectConfig projectConfig;
+    private OptimizelyConfig optimizelyConfig;
+
+    // Map containing variables list for every feature key used for merging variation and feature variables.
+    private Map<String, List<FeatureVariable>> featureKeyToVariablesMap;
```
jaeopt: We do not need to keep this instance. Can we remove it after initialization? Same for projectConfig, which is needed for construction only.

zashraf1985: remove as in? assign null? I beleive whole `OptimizelyConfigService` instance will be garbage collected as we are not holding any reference to the service.

jaeopt: To my understanding, your plan is to keep this OptimizelyConfigService instance for a cached copy of OptimizelyConfig.

fayyazarshad: No, we are only keeping OptimizelyConfig instance for cache. i.e
`private final AtomicReference<OptimizelyConfig> currentOptimizelyConfig = new AtomicReference<>();`

Moreover, I agree with removing featureKeyToVariablesMap as an instance variable and generate it directly in the method where it is being used like that.

`Map<String, List<FeatureVariable>> featureKeyToVariablesMap = generateFeatureKeyToVariablesMap();`

`List<FeatureVariable> featureVariables = featureKeyToVariablesMap.get(featureKey);`

On the other hand I think we can keep projectConfig as an instance variable because it is being used in multiple methods and it's a clean approach to not send it to each method as an arguement.

core-api/src/test/java/com/optimizely/ab/optimizelyconfig/OptimizelyConfigServiceTest.java
```diff
@@ -0,0 +1,514 @@
...
+import java.util.*;
+import static java.util.Arrays.asList;
+import static org.junit.Assert.*;
+
+@RunWith(MockitoJUnitRunner.class)
+public class OptimizelyConfigServiceTest {
+
+    private ProjectConfig projectConfig;
+    private OptimizelyConfigService optimizelyConfigService;
+    private OptimizelyConfig exptectedConfig;
```
mikecdavis: Nit, typo

core-api/src/test/java/com/optimizely/ab/optimizelyconfig/OptimizelyConfigServiceTest.java
```diff
@@ -0,0 +1,514 @@
...
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.mockito.runners.MockitoJUnitRunner;
+
+import java.util.*;
+import static java.util.Arrays.asList;
+import static org.junit.Assert.*;
+
+@RunWith(MockitoJUnitRunner.class)
```
mikecdavis: You can remove. Looks like you were going to mock the ProjectConfig.

fayyazarshad:
@jaeopt can you please review and approve, as its blocked upon your change request

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -1,5 +1,5 @@
 /****************************************************************************
- * Copyright 2016-2019, Optimizely, Inc. and contributors                   *
+ * Copyright 2020, Optimizely, Inc. and contributors                   *
```
zashraf1985: @fayyazarshad this should be 2016-2020

PR #352, opened 2019-12-20
feat: Added caching for OptimizelyConfig object

fayyazarshad:
## Summary
Updated the `OptimizelyConfig` implementation to always return cached object. `OptimizelyConfig` object will only be updated only when new `ProjectConfig` is received.

## Test Plan
Added Unit tests


core-api/src/main/java/com/optimizely/ab/config/PollingProjectConfigManager.java
```diff
@@ -1,6 +1,6 @@
 /**
  *
- *    Copyright 2019, Optimizely and contributors
+ *    Copyright 2020, Optimizely and contributors
```
zashraf1985: I would be 2019, 2020 or 2019-2020. copyright should cover all the years this file was modified in

core-api/src/test/java/com/optimizely/ab/config/PollingProjectConfigManagerTest.java
```diff
@@ -1,6 +1,6 @@
 /**
  *
- *    Copyright 2019, Optimizely and contributors
+ *    Copyright 2020, Optimizely and contributors
```
zashraf1985: same copyright comment applies here

PR #362, opened 2020-03-13
[HttpProjectConfigManager] Change "Fetching datafile from" log to debug level

Kapin:
## Summary
Change logging "Fetching data file from..." to debug level from info.

Info level logging on this is causing us two problems:

1) It's spamming our logs (which company wide are info) with this log line
2) We don't want the datafile name, which happens to correspond to the api key in our logs either

Debug solves both of these and we can optionally turn it on if we are seeing issues with datafile fetching.

## Test plan
N/A, only logging level

## Issues
N/A


Kapin:
@mikecdavis done

Kapin:
Would you also be able to retry the builds? The errors look unrelated to my change

mikeng13:
Builds are all passing now. Gonna merge.

PR #371, opened 2020-04-21
feature: add OptimizelyJSON

jaeopt:
## Summary
- Add OptimizelyJSON for JSON feature variable access


core-api/src/main/java/com/optimizely/ab/optimizelyjson/OptimizelyJSON.java
```diff
@@ -0,0 +1,129 @@
...
+            } else {
+                logger.error("Value for JSON key ({}) not found.", jsonKey);
+                break;
+            }
+        }
+
+        if (result == null) {
+            logger.error("Value for path could not be assigned to provided schema.");
+        }
+        return result;
```
pawels-optimizely, line 109: getValue should not return anything, it should just populate the object that you pass by the reference.

core-api/src/main/java/com/optimizely/ab/optimizelyjson/OptimizelyJSON.java
```diff
@@ -0,0 +1,153 @@
...
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+package com.optimizely.ab.optimizelyjson;
+
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
```
mikecdavis, line 20: We can't assume that `Gson` is available. I see two options:
1) We mimic the approach taken in [DefaultConfigParser](https://github.com/optimizely/java-sdk/blob/master/core-api/src/main/java/com/optimizely/ab/config/parser/DefaultConfigParser.java)
2) Take a stance on the serializer and include as a [shadow](https://imperceptiblethoughts.com/shadow/plugins/) dependency

Option 1 is the safer approach, since it has some precedence, but adds SDK complexity.
Option 2 is preferred but has a larger impact to the customer (my main concern is mobile). 

cc @mikeng13

core-api/src/test/java/com/optimizely/ab/optimizelyjson/OptimizelyJSONExtendedTest.java
```diff
@@ -0,0 +1,148 @@
...
+        MD1 md1 = oj1.getValue(null, MD1.class);
+        assertNotNull(md1);
+        assertEquals(md1.k1, "v1");
+        assertEquals(md1.k2, true);
+        assertEquals(md1.k3.kk1, 1.2, 0.01);
+        assertEquals(md1.k3.kk2.kkk1, true);
+        assertEquals((Double)md1.k3.kk2.kkk4[0], 5.7, 0.01);
+        assertEquals(md1.k3.kk2.kkk4[2], "vvv4");
+
+        // verify previous getValue does not destroy the data
```
pawels-optimizely, line 42: the purpose of this test is to call getValue twice with non empty json path on the same object. That will check if we don't modify the object by traversing and returning just a subset of the object.

core-api/src/test/java/com/optimizely/ab/optimizelyjson/OptimizelyJSONExtendedTest.java
```diff
@@ -0,0 +1,153 @@
...
+
+import com.optimizely.ab.config.parser.UnsupportedOperationException;
+import com.optimizely.ab.optimizelyjson.types.MD1;
+import com.optimizely.ab.optimizelyjson.types.MD2;
+import com.optimizely.ab.optimizelyjson.types.MD3;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+public class OptimizelyJSONExtendedTest extends OptimizelyJSONCoreTest {
```
mikecdavis, line 27: Instead of Core vs Extended you can use [assumeThat](https://junit.org/junit4/javadoc/4.12/org/junit/Assume.html#assumeThat(T,%20org.hamcrest.Matcher)) to conditionally ignore a parameterized variant.

core-api/src/test/java/com/optimizely/ab/optimizelyjson/OptimizelyJSONWithGsonParserTest.java
```diff
@@ -0,0 +1,103 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNull;
+
+public class OptimizelyJSONWithGsonParserTest extends OptimizelyJSONExtendedTest {
+    @Override
+    protected ConfigParser getParser() {
+        return new GsonConfigParser();
+    }
+
+    // Tests for GSON only
```
mikecdavis, line 38: I guess these classes are ok, but per the comments above, they should not be extending from `OptimizelyJSONExtendedTest`.

core-api/src/test/java/com/optimizely/ab/optimizelyjson/types/MD1.java
```diff
@@ -0,0 +1,23 @@
...
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+package com.optimizely.ab.optimizelyjson.types;
+
+public class MD1 {
```
mikecdavis, line 19: It's not clear what these test classes are supposed to represent. It might be clearer if you subclass these under something like `TestTypes.MD1` etc and add some doc strings.

core-api/src/main/java/com/optimizely/ab/config/parser/JsonHelpers.java
```diff
@@ -0,0 +1,81 @@
...
+ *    limitations under the License.
+ */
+package com.optimizely.ab.config.parser;
+
+import org.json.JSONArray;
+import org.json.JSONObject;
+
+import java.util.*;
+
+final class JsonHelpers {
```
mikecdavis, line 24: We should have a corresponding `JsonHelpersTest` class for unit tests.

core-api/src/main/java/com/optimizely/ab/config/parser/JacksonConfigParser.java
```diff
@@ -61,4 +63,23 @@ public ProjectConfigModule() {
             addDeserializer(Condition.class, new ConditionJacksonDeserializer(objectMapper));
         }
     }
+
+    @Override
+    public String toJson(Object src) throws ConfigParseException {
```
mikecdavis, line 68: We're missing unit tests on these new implementations.

PR #377, opened 2020-05-29
fix(log-level) Adjusting log level on audience evaluation logs

mnoman09:
## Summary
- Updating audience evaluation log level to be debug. The only info level log will be the one corresponding to the result of overall evaluation of an audience.


core-api/src/main/java/com/optimizely/ab/internal/ExperimentUtils.java
```diff
@@ -1,6 +1,6 @@
 /**
  *
- *    Copyright 2017-2019, Optimizely and contributors
+ *    Copyright 2017-2020, Optimizely and contributors
```
aliabbasrizvi, line 3: Why was this updated when nothing else is changing in this file

PR #378, opened 2020-06-12
feat: add support for authenticated datafile access

jaeopt:
## Summary
- add support for authenticated datafile access
- auth-token can be set by calling either
   (1) Optimizely optimizely = OptimizelyFactory.newDefaultInstance(String sdkKey, 
                                                                            String authDatafileToken, 
                                                                            String fallback); or
   (2) HttpProjectConfigManager.Builder builder = HttpProjectConfigManager.builder()
            .withSdkKey(sdkKey)
            .withAuthDatafileToken(authDatafileToken);
        Optimizely optimizely = OptimizelyFactory.newDefaultInstance(builder.build());

## Test plan
- test the target url and authenticated header are set appropriately both when auth-token is specified (authenticated datafile) or not (regular datafaile)

### DIFF
--- a/core-httpclient-impl/src/main/java/com/optimizely/ab/OptimizelyFactory.java
+++ b/core-httpclient-impl/src/main/java/com/optimizely/ab/OptimizelyFactory.java
@@ -179,6 +179,18 @@ public final class OptimizelyFactory {
      * @param fallback Fallback datafile string used by the ProjectConfigManager to be immediately available.
      */
     public static Optimizely newDefaultInstance(String sdkKey, String fallback) {
+        String datafileAuthToken = PropertyUtils.get(HttpProjectConfigManager.CONFIG_DATAFILE_AUTH_TOKEN);
+        return newDefaultInstance(sdkKey, fallback, datafileAuthToken);
+    }
+
+    /**
+     * Returns a new Optimizely instance with authenticated datafile support.
+     *
+     * @param sdkKey   SDK key used to build the ProjectConfigManager.
+     * @param fallback Fallback datafile string used by the ProjectConfigManager to be immediately available.
+     * @param datafileAuthToken  Token for authenticated datafile access.
+     */
+    public static Optimizely newDefaultInstance(String sdkKey, String fallback, String datafileAuthToken) {
         NotificationCenter notificationCenter = new NotificationCenter();
 
         HttpProjectConfigManager.Builder builder = HttpProjectConfigManager.builder()
@@ -186,24 +198,9 @@ public final class OptimizelyFactory {
             .withNotificationCenter(notificationCenter)
             .withSdkKey(sdkKey);
 
-        return newDefaultInstance(builder.build(), notificationCenter);
-    }
-
-    /**
-     * Returns a new Optimizely instance with authenticated datafile support.
-     *
-     * @param sdkKey   SDK key used to build the ProjectConfigManager.
-     * @param datafileAuthToken  Token for authenticated datafile access.
-     * @param fallback Fallback datafile string used by the ProjectConfigManager to be immediately available.
-     */
-    public static Optimizely newDefaultInstance(String sdkKey, String datafileAuthToken, String fallback) {
-        NotificationCenter notificationCenter = new NotificationCenter();
-
-        HttpProjectConfigManager.Builder builder = HttpProjectConfigManager.builder()
-            .withDatafile(fallback)
-            .withNotificationCenter(notificationCenter)
-            .withSdkKey(sdkKey)
-            .withDatafileAuthToken(datafileAuthToken);
+        if (datafileAuthToken != null) {
+            builder.withDatafileAuthToken(datafileAuthToken);
+        }
 
         return newDefaultInstance(builder.build(), notificationCenter);
     }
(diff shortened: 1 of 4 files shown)


core-httpclient-impl/src/main/java/com/optimizely/ab/OptimizelyFactory.java
```diff
@@ -189,6 +189,25 @@ public static Optimizely newDefaultInstance(String sdkKey, String fallback) {
...
     }
 
+    /**
+     * Returns a new Optimizely instance with authenticated datafile support.
+     *
+     * @param sdkKey   SDK key used to build the ProjectConfigManager.
+     * @param authDatafileToken  Token for authenticated datafile access.
+     * @param fallback Fallback datafile string used by the ProjectConfigManager to be immediately available.
+     */
+    public static Optimizely newDefaultInstance(String sdkKey, String authDatafileToken, String fallback) {
```
mjc1283, line 199: In other SDKs we named it `datafileAuthToken`, could you please follow that for all variable & method names?

core-httpclient-impl/src/main/java/com/optimizely/ab/config/HttpProjectConfigManager.java
```diff
@@ -125,14 +130,30 @@ protected ProjectConfig poll() {
         return null;
     }
 
+    HttpGet createHttpRequest() {
```
mjc1283, line 133: Should `createHttpRequest` be `private`?

core-httpclient-impl/src/main/java/com/optimizely/ab/config/HttpProjectConfigManager.java
```diff
@@ -125,14 +130,30 @@ protected ProjectConfig poll() {
...
     public static Builder builder() {
         return new Builder();
     }
 
     public static class Builder {
         private String datafile;
         private String url;
+        private String authDatafileToken = null;
         private String format = "https://cdn.optimizely.com/datafiles/%s.json";
+        private String authFormat = "https://www.optimizely-cdn.com/datafiles/auth/%s.json";
```
mjc1283, line 156: The domain is `config.optimizely.com`:
```suggestion
        private String authFormat = "https://config.optimizely.com/datafiles/auth/%s.json";
```

core-httpclient-impl/src/main/java/com/optimizely/ab/config/HttpProjectConfigManager.java
jaeopt: Private must be right one. This default access will be useful for header testing.

mjc1283: Then let's use the [VisibleForTesting annotation](https://github.com/optimizely/java-sdk/blob/89d18ac95bf71f0942d89e6363785d3ea5f79b7b/core-api/src/main/java/com/optimizely/ab/annotations/VisibleForTesting.java#L22).

core-httpclient-impl/src/main/java/com/optimizely/ab/OptimizelyFactory.java
```diff
@@ -189,6 +189,25 @@ public static Optimizely newDefaultInstance(String sdkKey, String fallback) {
...
     }
 
+    /**
+     * Returns a new Optimizely instance with authenticated datafile support.
+     *
+     * @param sdkKey   SDK key used to build the ProjectConfigManager.
+     * @param datafileAuthToken  Token for authenticated datafile access.
+     * @param fallback Fallback datafile string used by the ProjectConfigManager to be immediately available.
+     */
+    public static Optimizely newDefaultInstance(String sdkKey, String datafileAuthToken, String fallback) {
```
mikecdavis, line 199: The cascading signatures here don't line up nicely. We should consistently call the next method in the chain.
```java
public static Optimizely newDefaultInstance(String sdkKey) {
    ...
    return newDefaultInstance(sdkKey, null)
}

public static Optimizely newDefaultInstance(String sdkKey, String fallback) {
    ...
    String datafileAuthToken = PropertyUtils.get(HttpProjectConfigManager.CONFIG_SDK_AUTH_TOKEN);
    return newDefaultInstance(sdkKey, fallback, datafileAuthToken)
}

public static Optimizely newDefaultInstance(String sdkKey, String fallback, String datafileAuthToken) {
    ...
    return newDefaultInstance(builder.build(), notificationCenter);
}
```
Unfortunately these are all strings so order needs to be paid attention to. We should also provide a configuration option for `datafileAuthToken` so customers can deploy this via `optimizely.properties` and not have explicitly provide this value to the factory method. See the example above.

core-httpclient-impl/src/main/java/com/optimizely/ab/OptimizelyFactory.java
```diff
@@ -179,13 +179,29 @@ public static Optimizely newDefaultInstance(String sdkKey) {
      * @param fallback Fallback datafile string used by the ProjectConfigManager to be immediately available.
      */
     public static Optimizely newDefaultInstance(String sdkKey, String fallback) {
+        String datafileAuthToken = PropertyUtils.get(HttpProjectConfigManager.CONFIG_DATAFILE_AUTH_TOKEN);
```
mikeng13, line 182: We've been using `datafileAccessToken` in the other SDKs. Let's stick with the same name throughout.

PR #388, opened 2020-08-03
docs: changed reference from the wrong classname to the correct one i…

umermansoor:
…n Javadocs

## Summary
- Fixed Javadocs in the `OptimizelyFactory` class that were referencing to (non-existent) `OptimizelyClients`

Why I changed it? Because it is confusion for users

## Issues
- "THING-1234" or "Fixes #123"


umermansoor:
Hi, is there an issue with the PR?

PR #390, opened 2020-08-07
Add MatchRegistry for custom match implementations.

mikecdavis:
## Summary
- Replace `MatchType` in favor of `MatchRegistry`
- Add conditionValue to `Match#match` interface.
- Remove the abstract `AttributeMatch` class.
- Remove `ExactNumberMatch` class.
- Add NumberComparator for standardizing Number comparisons
- Refactor Semantic Versioning to use static convenience method

Incorporating a public MatchRegistry will allow consumers to provide their own implementations of the Match interface.


mikecdavis:
Currently failing a single scenario in the [FSC](https://github.com/optimizely/fullstack-sdk-compatibility-suite/blob/7459fec6ad23063ff476e8e42fb3f0d7aae9b99e/features/activate/audience_targeting/typed_audiences.feature#L210):

core-api/src/main/java/com/optimizely/ab/config/audience/match/GTMatch.java
```diff
@@ -18,24 +18,13 @@
...
-            }
-        } catch (Exception e) {
-            return null;
-        }
-        return null;
+    public Boolean eval(Object conditionValue, Object attributeValue) throws UnknownValueTypeException {
+        return NumberComparator.compare(attributeValue, conditionValue) > 0;
     }
-}
+}
```
msohailhussain, line 30: eol

core-api/src/main/java/com/optimizely/ab/config/audience/match/MatchRegistry.java
```diff
@@ -0,0 +1,83 @@
...
+ * All supported Match implementations must be registed with this registry.
+ * Third-party {@link Match} implementations may also be registered to provide
+ * additional functionality.
+ */
+public class MatchRegistry {
+
+    private static final Map<String, Match> registry = new ConcurrentHashMap<>();
+    public static final String EXISTS = "exists";
+    public static final String EXACT = "exact";
+    public static final String GREATER_THAN = "gt";
```
msohailhussain, line 33: nit: alphabetize.

core-api/src/main/java/com/optimizely/ab/config/audience/match/NumberComparator.java
```diff
@@ -0,0 +1,37 @@
...
+
+import static com.optimizely.ab.internal.AttributesUtil.isValidNumber;
+
+public class NumberComparator {
+    public static int compare(Object o1, Object o2) throws UnknownValueTypeException {
+        if (!isValidNumber(o1)) {
+            throw new UnknownValueTypeException();
+        }
+
+        if (!isValidNumber(o2)) {
```
msohailhussain, line 31: Can we merge into the above `if` condition?

core-api/src/main/java/com/optimizely/ab/config/audience/match/NumberComparator.java
```diff
@@ -0,0 +1,37 @@
...
+            throw new UnknownValueTypeException();
+        }
+
+        return compareUnsafe(o1, o2);
+    }
+
+    public static int compareUnsafe(Object o1, Object o2) {
```
msohailhussain: doc string will be helpful here.

core-api/src/main/java/com/optimizely/ab/config/audience/match/SubstringMatch.java
```diff
@@ -18,11 +18,9 @@
 
 import javax.annotation.Nullable;
 
-class SubstringMatch extends AttributeMatch<String> {
-    String value;
+class SubstringMatch implements Match {
```
msohailhussain, line 25: update Copyright year.

msohailhussain (approved):
lgtm. few nits

core-api/src/main/java/com/optimizely/ab/config/audience/match/Match.java
```diff
@@ -20,5 +20,5 @@
 
 public interface Match {
     @Nullable
-    Boolean eval(Object attributeValue);
+    Boolean eval(Object conditionValue, Object attributeValue) throws UnexpectedValueTypeException, UnknownValueTypeException;
```
jaeopt, line 23: Do we need separate exception types for UnexpectedValueTypeException and UnknownValueTypeException?
They're similar and not much gain for additional complexity to differentiate them.

jaeopt (changes requested):
It looks great. One change suggested.

core-api/src/main/java/com/optimizely/ab/config/audience/match/Match.java
mikecdavis: This is for backwards compatibility, albeit I can update the names perhaps. The difference is in the logging behavior. One error is communicated as an SDK compatibility issue warning to upgrade on the condition value, while the other is saying that the attribute value type is unknown.

mikecdavis: The distinction is since the conditionValue comes from the datafile, so if a value is of an "unexpected" type then its likely the result of an out of date SDK. Since the attributeValue is provided at runtime via the user attributes, then we log that the type is "unknown".

jaeopt: Got it. We can consider adding some clarifications.

jaeopt (approved):

core-api/src/main/java/com/optimizely/ab/config/audience/match/Match.java
mikecdavis: I added doc strings for all the exception classes in this package.

PR #394, opened 2020-09-02
fix: Semantic version fix and multiple plus signs check added

msohailhussain:
## Summary
- Multiple fixed for sem-versioning
- Compare build version and pre-release fix
- Multiple `+` added.

## Test plan
All FSC tests must pass
Added unit tests


core-api/src/main/java/com/optimizely/ab/config/audience/match/SemanticVersion.java
```diff
@@ -93,6 +109,17 @@ private int dotCount(String prefixVersion) {
         return count;
     }
 
+    private boolean isbuildCountMultiple() {
```
msohailhussain, line 112: instead, just say isValidBuildMetadata?

thomaszurkan-optimizely (approved):
Looks Good!

PR #400, opened 2020-09-21
[RFC] Add encapsulating UserContext.

mikecdavis:
## Summary
- Straw man approach for UserContext to support new decide API
- UserContext hold reference to Optimizely client.
- Stop returning `null` from OptimizelyJSON
- Update Example.java to exemplify the suggested approach

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/Optimizely.java
+++ b/core-api/src/main/java/com/optimizely/ab/Optimizely.java
@@ -1141,6 +1141,10 @@ public class Optimizely implements AutoCloseable {
         return notificationCenter.addNotificationHandler(clazz, handler);
     }
 
+    public UserContext CreateUserContext(String userId) {
+        return new UserContext(userId, this);
+    }
+
     //======== Builder ========//
 
     /**
--- a/core-api/src/main/java/com/optimizely/ab/UserContext.java
+++ b/core-api/src/main/java/com/optimizely/ab/UserContext.java
@@ -1,0 +1,39 @@
+package com.optimizely.ab;
+
+import com.optimizely.ab.optimizelyjson.OptimizelyJSON;
+
+import java.util.Collections;
+import java.util.HashMap;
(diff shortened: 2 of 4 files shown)


core-api/src/main/java/com/optimizely/ab/UserContext.java
```diff
@@ -0,0 +1,39 @@
...
+        this.optimizely = optimizely;
+        this.userAttributes = userAttributes;
+    }
+
+    public void setAttribute(String name, Object value) {
+        this.userAttributes.put(name, value);
+    }
+
+    public OptimizelyJSON decide(String decideKey) {
+       return optimizely.getAllFeatureVariables(decideKey, userId, userAttributes);
```
mikecdavis, line 29: Note, I'm using the existing `getAllFeatureVariables` method, but this would probably be replaced for a top level `decide` method defined on Optimizely.

core-api/src/main/java/com/optimizely/ab/UserContext.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    public void track(String eventName) {
```
aliabbasrizvi, line 32: I think we picked `trackEvent` later on.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -1141,6 +1141,10 @@ public int addLogEventNotificationHandler(NotificationHandler<LogEvent> handler)
         return notificationCenter.addNotificationHandler(clazz, handler);
     }
 
+    public UserContext CreateUserContext(String userId) {
```
aliabbasrizvi, line 1144: Shouldn't this be `createUserContext`?

aliabbasrizvi (commented):
Seems to be trending in the right direction.

jaeopt:
Done its purpose!

PR #402, opened 2020-09-23
Fix: logging issue in quick-start application

mnoman09:
## Summary
- Replaced slf4j-log4j12 lib with slf4j-simple due to vulnerability issue in log4j 1.2 version
- Also this PR will resolve the logging issue in quick-start application

## Test plan
FSC and all unit tests should pass.

### DIFF
--- a/java-quickstart/build.gradle
+++ b/java-quickstart/build.gradle
@@ -4,7 +4,8 @@ dependencies {
 
     compile group: 'com.google.code.gson', name: 'gson', version: '2.8.6'
     compile group: 'org.apache.httpcomponents', name: 'httpclient', version: '4.5.12'
-    compile group: 'org.apache.logging.log4j', name: 'log4j-slf4j-impl', version: '2.13.3'
+    compile group: 'org.slf4j', name: 'slf4j-simple', version: '1.7.30'
+
     testCompile group: 'junit', name: 'junit', version: '4.12'
 }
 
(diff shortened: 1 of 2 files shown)


java-quickstart/build.gradle
```diff
@@ -4,7 +4,8 @@ dependencies {
 
     compile group: 'com.google.code.gson', name: 'gson', version: '2.8.6'
     compile group: 'org.apache.httpcomponents', name: 'httpclient', version: '4.5.12'
-    compile group: 'org.apache.logging.log4j', name: 'log4j-slf4j-impl', version: '2.13.3'
+    compile group: 'org.slf4j', name: 'slf4j-simple', version: '1.7.30'
```
Mat001, line 7: I see this is a revert to the line that was there previously. That works, tested it. But I'm not sure if maybe new Apache logger might have different logging properties setup, and this line didn't need to be reverted and instead Apache logger might need a different setup. I'm not sure.

Mat001 (approved):

java-quickstart/build.gradle
mnoman09: No actually previously it was     _compile group: 'org.slf4j', name: 'slf4j-log4j12', version: '1.7.30'_  which was causing vulnerability issue due to transitive property of log4j. I now replaced it with alternative library  slf4j-simple.

aliabbasrizvi (approved):

PR #406, opened 2020-10-09
feat(decide): add a new set of decide apis

jaeopt:
## Summary
Add a new set of Decide APIs:
- define OptimizelyUserContext/OptimizelyDecision/OptimizelyDecideOption
- add createUserContext API to Optimizely
- add defaultDecideOption to Optimizely constructor and builder
- decide/decideAll/trackEvent


core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -1190,6 +1219,7 @@ public static Builder builder() {
         private OptimizelyConfigManager optimizelyConfigManager;
         private UserProfileService userProfileService;
         private NotificationCenter notificationCenter;
+        private OptimizelyDecideOption[] defaultDecideOptions;
```
mikecdavis, line 1222: Consider `List<OptimizelyDecideOption>` per previous comment.

mikecdavis (changes requested):
The required change is to the documentation. The others are strong recommendations to flatten the package and to use a `List` as opposed to an array.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -1261,6 +1455,11 @@ public Builder withDatafile(String datafile) {
             return this;
         }
 
+        public Builder withDefaultDecideOptions(List<OptimizelyDecideOption> options) {
+            this.defaultDecideOptions = Collections.unmodifiableList(options);
```
mnoman09, line 1459: I think we can revise this and instead of converting options to unmodifiableList here we can just assign it and in build method we can check if options is not null, then change it to unmodifiableList because it will cause null pointer exception.

mnoman09 (changes requested):

core-api/src/main/java/com/optimizely/ab/config/audience/Condition.java
```diff
@@ -17,15 +17,25 @@
...
 public interface Condition<T> {
 
+    @Nullable
+    Boolean evaluate(ProjectConfig config,
+                     Map<String, ?> attributes,
+                     List<OptimizelyDecideOption> options,
+                     DecisionReasons reasons);
+
     @Nullable
     Boolean evaluate(ProjectConfig config, Map<String, ?> attributes);
```
mikecdavis, line 39: Do we need to keep this signature? Feels like we only need it for testing. Since "we" own the calling of this method we can pass defaults there.

core-api/src/main/java/com/optimizely/ab/optimizelydecision/DecisionReasons.java
```diff
@@ -0,0 +1,60 @@
...
+        errors.add(message);
+    }
+
+    public void addInfo(String message) {
+        logs.add(message);
+    }
+
+    public String addInfoF(String format, Object... args) {
+        String message = String.format(format, args);
+        if(includeReasons) addInfo(message);
```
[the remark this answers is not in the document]
yasirfolio3: `addInfo` doesn't check for `includeReasons` whereas addInfof does, any reason behind that?

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -1086,6 +1092,193 @@ public OptimizelyConfig getOptimizelyConfig() {
...
+    public OptimizelyUserContext createUserContext(@Nonnull String userId,
+                                                   @Nonnull Map<String, ?> attributes) {
+        return new OptimizelyUserContext(this, userId, attributes);
+    }
+
+    public OptimizelyUserContext createUserContext(@Nonnull String userId) {
+        return new OptimizelyUserContext(this, userId);
+    }
+
+    OptimizelyDecision decide(@Nonnull OptimizelyUserContext user,
```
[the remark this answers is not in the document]
jaeopt: I agree. We better keep "user" and I hope we eventually make them all consistent. But for legacy APIs to be deprecated, we keep them as is.

core-api/src/main/java/com/optimizely/ab/optimizelydecision/DecisionReasons.java
```diff
@@ -0,0 +1,60 @@
...
+
+    public void addError(String message) {
+        errors.add(message);
+    }
+
+    public void addInfo(String message) {
+        logs.add(message);
+    }
+
+    public String addInfoF(String format, Object... args) {
```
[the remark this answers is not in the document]
jaeopt: I found that we do not keep "addInfo" and "addInfof" both. I merge them as a single "addInfo", which looks cleaner.

core-api/src/main/java/com/optimizely/ab/optimizelydecision/DecisionReasons.java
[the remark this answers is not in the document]
jaeopt: I merged them to avoid confusion. See comments above.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -104,7 +109,8 @@ private Optimizely(@Nonnull EventHandler eventHandler,
                        @Nullable UserProfileService userProfileService,
                        @Nonnull ProjectConfigManager projectConfigManager,
                        @Nullable OptimizelyConfigManager optimizelyConfigManager,
-                       @Nonnull NotificationCenter notificationCenter
+                       @Nonnull NotificationCenter notificationCenter,
+                       @Nonnull List<OptimizelyDecideOption> defaultDecideOptions
```
msohailhussain, line 113: can you add an overloaded method and pass `[]`  without defaultDecideOptions?

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -114,6 +120,7 @@ private Optimizely(@Nonnull EventHandler eventHandler,
         this.projectConfigManager = projectConfigManager;
         this.optimizelyConfigManager = optimizelyConfigManager;
         this.notificationCenter = notificationCenter;
+        this.defaultDecideOptions = defaultDecideOptions;
```
msohailhussain, line 123: we should clone it.

core-api/src/main/java/com/optimizely/ab/OptimizelyUserContext.java
```diff
@@ -0,0 +1,186 @@
...
+        return optimizely;
+    }
+
+    /**
+     * Set an attribute for a given key.
+     *
+     * @param key An attribute key
+     * @param value An attribute value
+     */
+    public void setAttribute(@Nonnull String key, @Nonnull Object value) {
```
msohailhussain, line 73: we may pass `null`

core-api/src/main/java/com/optimizely/ab/OptimizelyUserContext.java
```diff
@@ -0,0 +1,186 @@
...
+        return optimizely.decide(this, key, options);
+    }
+
+    /**
+     * Returns a decision result ({@link OptimizelyDecision}) for a given flag key and a user context, which contains all data required to deliver the flag.
+     *
+     * @param key A flag key for which a decision will be made.
+     * @return A decision result.
+     */
+    public OptimizelyDecision decide(String key) {
```
msohailhussain, line 97: Nonnull.

core-api/src/main/java/com/optimizely/ab/OptimizelyUserContext.java
```diff
@@ -0,0 +1,186 @@
...
+        return optimizely.decide(this, key, Collections.emptyList());
```
msohailhussain, line 98: change it to 
`this.decide(key, Collections.emptyList())`

core-api/src/main/java/com/optimizely/ab/OptimizelyUserContext.java
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    /**
+     * Returns a key-map of decision results for multiple flag keys and a user context.
+     *
+     * @param keys A list of flag keys for which decisions will be made.
+     * @return All decision results mapped by flag keys.
+     */
+    public Map<String, OptimizelyDecision> decideForKeys(@Nonnull List<String> keys) {
+        return optimizely.decideForKeys(this, keys, Collections.emptyList());
```
msohailhussain, line 123: call decideForKeys of this class.

core-api/src/main/java/com/optimizely/ab/OptimizelyUserContext.java
```diff
@@ -0,0 +1,186 @@
...
+        return optimizely.decideAll(this, options);
+    }
+
+    /**
+     * Returns a key-map of decision results ({@link OptimizelyDecision}) for all active flag keys.
+     *
+     * @return A dictionary of all decision results, mapped by flag keys.
+     */
+    public Map<String, OptimizelyDecision> decideAll() {
+        return optimizely.decideAll(this, Collections.emptyList());
```
msohailhussain, line 142: same as above.

core-api/src/main/java/com/optimizely/ab/bucketing/Bucketer.java
```diff
@@ -123,42 +128,56 @@ private Variation bucketToVariation(@Nonnull Experiment experiment,
...
+     * @param options            An array of decision options
+     * @param reasons            Decision log messages
      * @return Variation the user is bucketed into or null.
      */
     @Nullable
     public Variation bucket(@Nonnull Experiment experiment,
                             @Nonnull String bucketingId,
-                            @Nonnull ProjectConfig projectConfig) {
+                            @Nonnull ProjectConfig projectConfig,
+                            @Nonnull List<OptimizelyDecideOption> options,
```
msohailhussain, line 140: am not sure, if we need options in bucketing logic.

core-api/src/main/java/com/optimizely/ab/bucketing/DecisionService.java
```diff
@@ -81,24 +83,28 @@ public DecisionService(@Nonnull Bucketer bucketer,
...
     public Variation getVariation(@Nonnull Experiment experiment,
                                   @Nonnull String userId,
                                   @Nonnull Map<String, ?> filteredAttributes,
-                                  @Nonnull ProjectConfig projectConfig) {
-
-        if (!ExperimentUtils.isExperimentActive(experiment)) {
+                                  @Nonnull ProjectConfig projectConfig,
+                                  @Nonnull List<OptimizelyDecideOption> options,
+                                  @Nonnull DecisionReasons reasons) {
+        if (!ExperimentUtils.isExperimentActive(experiment, options, reasons)) {
```
msohailhussain, line 98: options not needed.

core-api/src/main/java/com/optimizely/ab/bucketing/DecisionService.java
```diff
@@ -198,94 +233,133 @@ public FeatureDecision getVariationForFeature(@Nonnull FeatureFlag featureFlag,
...
      */
     @Nullable
     Variation getStoredVariation(@Nonnull Experiment experiment,
                                  @Nonnull UserProfile userProfile,
-                                 @Nonnull ProjectConfig projectConfig) {
+                                 @Nonnull ProjectConfig projectConfig,
+                                 @Nonnull List<OptimizelyDecideOption> options,
+                                 @Nonnull DecisionReasons reasons) {
+
+        if (options.contains(OptimizelyDecideOption.IGNORE_USER_PROFILE_SERVICE)) return null;
```
msohailhussain, line 361: Can't we make all this Variation method independent of options? And from the root level, we should decide?

msohailhussain (changes requested):
will review some more classes in a detail later.

core-api/src/main/java/com/optimizely/ab/optimizelydecision/OptimizelyDecision.java
```diff
@@ -0,0 +1,148 @@
...
+    private final String flagKey;
+
+    @Nonnull
+    private final OptimizelyUserContext userContext;
+
+    @Nonnull
+    private List<String> reasons;
+
+
+    public OptimizelyDecision(@Nullable String variationKey,
```
msohailhussain, line 50: I will prefer the following sequence for args.  Just a suggestion.
1. userContext
2. flagKey
3. rule_key
4. variation_key
5. enabled (It may come after flagKey as well.)
6. reasons

core-api/src/main/java/com/optimizely/ab/optimizelydecision/OptimizelyDecision.java
```diff
@@ -0,0 +1,148 @@
...
+    public OptimizelyUserContext getUserContext() {
+        return userContext;
+    }
+
+    @Nonnull
+    public List<String> getReasons() {
+        return reasons;
+    }
+
+    public static OptimizelyDecision createErrorDecision(@Nonnull String key,
```
msohailhussain, line 100: `newErrorDecision`

core-api/src/main/java/com/optimizely/ab/optimizelydecision/OptimizelyDecision.java
```diff
@@ -0,0 +1,148 @@
...
+            null,
+            false,
+            new OptimizelyJSON(Collections.emptyMap()),
+            null,
+            key,
+            user,
+            Arrays.asList(error));
+    }
+
+    public boolean hasFailed() {
```
msohailhussain, line 113: what's the purpose of this method?

msohailhussain (commented):
OK, i have reviewed all logics, need to review unit tests which I may take a look later.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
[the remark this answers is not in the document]
mikecdavis: Hm I see, you have this as package private. My suggestion would not require that and still provide access to this functionality.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
mikecdavis: @msohailhussain can you elaborate? this is a private constructor any defaults should be handled in the builder. I don't think we should add another constructor.

core-api/src/main/java/com/optimizely/ab/OptimizelyUserContext.java
mikecdavis: I guess it technically won't hurt anything, but it won't be useful in any way to have a null key

mikecdavis (commented):
I agree with much of @msohailhussain concerns. it was difficult to reason why we need to pass options and reasons so far down the stack and begs the question if there a better way to encapsulate that. For example providing a richer response object would be a cleaner alternative for capturing reasons.

core-api/src/main/java/com/optimizely/ab/OptimizelyUserContext.java
msohailhussain: am asking for `Object` value, it can be `null`.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
msohailhussain: Yeah i missed private, this is fine.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
jaeopt: It's cloned already in builder, so no need for it here.

core-api/src/main/java/com/optimizely/ab/bucketing/Bucketer.java
jaeopt: Not used for now, but will keep this consistently for future options.

core-api/src/main/java/com/optimizely/ab/bucketing/DecisionService.java
jaeopt: Not used for now, but will keep this consistently for future options.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -1086,6 +1092,192 @@ public OptimizelyConfig getOptimizelyConfig() {
...
+        if (projectConfig == null) {
+            return OptimizelyDecision.createErrorDecision(key, user, DecisionMessage.SDK_NOT_READY.reason());
+        }
+
+        FeatureFlag flag = projectConfig.getFeatureKeyMapping().get(key);
+        if (flag == null) {
+            return OptimizelyDecision.createErrorDecision(key, user, DecisionMessage.FLAG_KEY_INVALID.reason(key));
+        }
+
+        String userId = user.getUserId();
```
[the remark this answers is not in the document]
jaeopt: No need since this private is called thru a valid user context

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -1086,6 +1092,192 @@ public OptimizelyConfig getOptimizelyConfig() {
...
+        }
+
+        if (keys.isEmpty()) return decisionMap;
+
+        List<OptimizelyDecideOption> allOptions = getAllOptions(options);
+
+        for (String key : keys) {
+            OptimizelyDecision decision = decide(user, key, options);
+            if (!allOptions.contains(OptimizelyDecideOption.ENABLED_FLAGS_ONLY) || decision.getEnabled()) {
+                decisionMap.put(key, decision);
```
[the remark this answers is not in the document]
jaeopt: Not sure about this comment. Clarification?

core-api/src/main/java/com/optimizely/ab/OptimizelyUserContext.java
jaeopt: Will keep it to save one-step indirection.

core-api/src/main/java/com/optimizely/ab/OptimizelyUserContext.java
jaeopt: Will keep it to save one-step indirection.

core-api/src/main/java/com/optimizely/ab/OptimizelyUserContext.java
jaeopt: Will keep it to save one-step indirection.

core-api/src/main/java/com/optimizely/ab/optimizelydecision/OptimizelyDecision.java
jaeopt: It makes sense in general. Here, I intended with the current order for more important decision info first and then debug/info values following.

core-api/src/main/java/com/optimizely/ab/optimizelydecision/OptimizelyDecision.java
jaeopt: removed!

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -1126,6 +1132,200 @@ public OptimizelyConfig getOptimizelyConfig() {
...
+        List<FeatureFlag> allFlags = projectConfig.getFeatureFlags();
+        List<String> allFlagKeys = new ArrayList<>();
+        for (int i = 0; i < allFlags.size(); i++) allFlagKeys.add(allFlags.get(i).getKey());
+
+        return decideForKeys(user, allFlagKeys, options);
+    }
+
+    private List<OptimizelyDecideOption> getAllOptions(List<OptimizelyDecideOption> options) {
+        List<OptimizelyDecideOption> copiedOptions = new ArrayList(defaultDecideOptions);
+        copiedOptions.addAll(options);
```
mnoman09, line 1298: As it would throw null pointer exception if options is null. 
```suggestion
        if (options != null) {
            copiedOptions.addAll(options);
        }
```

core-api/src/main/java/com/optimizely/ab/OptimizelyUserContext.java
```diff
@@ -0,0 +1,195 @@
...
+    private final Optimizely optimizely;
+
+    private static final Logger logger = LoggerFactory.getLogger(OptimizelyUserContext.class);
+
+    public OptimizelyUserContext(@Nonnull Optimizely optimizely,
+                                 @Nonnull String userId,
+                                 @Nonnull Map<String, Object> attributes) {
+        this.optimizely = optimizely;
+        this.userId = userId;
+        this.attributes = Collections.synchronizedMap(new HashMap<>(attributes));
```
mnoman09, line 48: This will also throw null pointer exception if attributes value is null. Plus we need to initialize it as new empty hashMap because of set attributes method as it is assuming that attributes is not null.
```suggestion
        if (attributes != null) {
            this.attributes = Collections.synchronizedMap(new HashMap<>(attributes));
        } else {
            this.attributes = Collections.synchronizedMap(new HashMap<>());
        }
```

core-api/src/main/java/com/optimizely/ab/OptimizelyUserContext.java
```diff
@@ -0,0 +1,195 @@
...
+    public OptimizelyUserContext(@Nonnull Optimizely optimizely, @Nonnull String userId) {
+        this(optimizely, userId, Collections.EMPTY_MAP);
+    }
+
+    public String getUserId() {
+        return userId;
+    }
+
+    public Map<String, Object> getAttributes() {
+        return new HashMap<String, Object>(attributes);
```
mnoman09, line 60: I think It's not needed anymore.
```suggestion
        return attributes;
```

mnoman09 (changes requested):
Overall Looks good. Pointed out some changes which might cause null pointer exception.

core-api/src/main/java/com/optimizely/ab/OptimizelyUserContext.java
jaeopt: This copy will be necessary to provide a snapshot of attributes since it can be updated in the middle of decision process.

mnoman09: can you specify a particular scenario in which It would be useful as currently I am using my branch of user-context-temp without this new Hashmap and it's working fine. If you can specify the scenario in which there would be a difference then we can add that test in FSC to verify.

jaeopt: The user attributes can be changed concurrently while decide api calls are being processed. Without the copy, attributes values may change while processing. Not sure if we need to protect this with FSC :)

core-api/src/main/java/com/optimizely/ab/OptimizelyUserContext.java
mikecdavis: nit, but I agree with @msohailhussain . best to mitigate divergence within a class first. Also since Optimizely#Decide is private it doesn't need these overloaded signatures and they are purely a convenience on _this_ class.

mikecdavis (approved):

msohailhussain (approved):
lgtm

mnoman09:
Lgtm

mnoman09 (approved):

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -1126,6 +1129,200 @@ public OptimizelyConfig getOptimizelyConfig() {
...
+        DecisionReasons decisionReasons = DefaultDecisionReasons.newInstance(allOptions);
+
+        Map<String, ?> copiedAttributes = new HashMap<>(attributes);
+        FeatureDecision flagDecision = decisionService.getVariationForFeature(
+            flag,
+            userId,
+            copiedAttributes,
+            projectConfig,
+            allOptions,
+            decisionReasons);
```
zashraf1985, line 1184: It appears that `decisionReasons` is expected to be modified from inside. In my experience, this kind of flow gets a little confusing when someone is following the code flow. On suggestion can be to return `Pair<Decision, Reasons>` from`getVariationForFeature`. This will clearly tell that decision reasons are also coming back from this function and original object will also be intact causing no confusion. what do you think

jaeopt: A good point! We also consider that option as well. It's dropped since it's a bit complicated to process and merge (and/or audience, etc) at each stage. We can find better solutions. I think you can try with other sdks with your suggestion and I can come back to this if we see a good solution.

thomaszurkan-optimizely (changes requested):
Please see my comments below.  It seems like a bad pattern to be passing around a parameter that is constantly being augmented.  We should not propagate this pattern to all SDKs. And a nit to remove the star for including decision package.

jaeopt:
@thomaszurkan-optimizely I'm wondering why you think it's not thread-safe. Each decide-api call in a thread will keep track of its own reason instance, so should be thread-safe.
As discussed above, I'm also wondering if we have any better solutions and keep it open. I'll follow up other SDKs alternative solutions and get back to fix if needed.

PR #407, opened 2020-10-09
feat(decide): add decide API

jaeopt:
## Summary
The second part of 2 PRs for Decide API:
- add decide/decideAll/trackEvent API implementation
- fix DecisionService and Bucketer to support DecideOptions and DecisionReasons (this fix should not break existing custom DecisionServerice/Bucketer)


core-api/src/main/java/com/optimizely/ab/optimizelyusercontext/OptimizelyUserContext.java
```diff
@@ -53,44 +68,284 @@ public Optimizely getOptimizely() {
...
+        return allOptions;
+    }
+
+    public static String getFlagKeyInvalidMessage(String flagKey) {
+        return String.format(FLAG_KEY_INVALID, flagKey);
+    }
+
+    public static String getVariableValueInvalidMessage(String variableKey) {
+        return String.format(VARIABLE_VALUE_INVALID, variableKey);
+    }
```
mikecdavis, line 305: One-liner "helper" methods then to add more cognitive overhead than they alleviate. Consider in-lining the String.format directly. Alternatively if you wanted to provide some abstraction, then a customer enum class can be used to define the format functionality:

```java
    private enum Messages {
        FLAG_KEY_INVALID("Flag key \"%s\" is not in datafile."),
        VARIABLE_VALUE_INVALID("Variable value for key \"%s\" is invalid or wrong type.");

        private final String format;

        Messages(String format) {
            this.format = format;
        }

        public String format(Object...args) {
            return String.format(format, args);
        }
    }
```
Which can then be consumed as:
```java
Messages.FLAG_KEY_INVALID.format("key");
```
Note that this implementation reads more direct about what is happening. We're formatting a predefined message.

mikecdavis (changes requested):

PR #410, opened 2020-11-13
fix: decision metadata enabled fix.

msohailhussain:
## Summary
- IsFeatureEnabled was sending impression before finalizing all checks of feature enabled. Just made it after finalization.

## Test plan
- can't add unit tests. 
- only FSC should be passed.


jaeopt (approved):
LGTM

PR #414, opened 2020-12-15
chore: fix sample codes to use decide apis

jaeopt:
## Summary
- Fix sample codes from legacy APIs to new decide APIs


msohailhussain (approved):
lgtm

PR #415, opened 2020-12-18
fix(decide): change to return reasons as a part of tuple in decision hierarchy

jaeopt:
## Summary
- Change to return reasons as a part of tuple in decision hierarchy (from passing down a reference for modification)

## Test plan
- Validate that reasons are collected properly with or without INCLUDE_REASONS option.


core-api/src/main/java/com/optimizely/ab/bucketing/DecisionService.java
```diff
@@ -344,15 +368,14 @@ Variation getWhitelistedVariation(@Nonnull Experiment experiment,
      * @param experiment  {@link Experiment} in which the user was bucketed.
      * @param userProfile {@link UserProfile} of the user.
      * @param projectConfig      The current projectConfig
-     * @param reasons            Decision log messages
      * @return null if the {@link UserProfileService} implementation is null or the user was not previously bucketed.
```
yavorona, line 371: Should update comments with DecisionResponse

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -427,7 +423,7 @@ private Boolean isFeatureEnabled(@Nonnull ProjectConfig projectConfig,
 
         Map<String, ?> copiedAttributes = copyAttributes(attributes);
         FeatureDecision.DecisionSource decisionSource = FeatureDecision.DecisionSource.ROLLOUT;
-        FeatureDecision featureDecision = decisionService.getVariationForFeature(featureFlag, userId, copiedAttributes, projectConfig);
+        FeatureDecision featureDecision = decisionService.getVariationForFeature(featureFlag, userId, copiedAttributes, projectConfig).getResult();
```
thomaszurkan-optimizely, line 426: The decision service methods are all marked as nullable return value.

core-api/src/main/java/com/optimizely/ab/bucketing/DecisionService.java
```diff
@@ -136,42 +143,49 @@ public Variation getVariation(@Nonnull Experiment experiment,
...
+        return new DecisionResponse(null, reasons);
     }
 
     @Nullable
-    public Variation getVariation(@Nonnull Experiment experiment,
-                                  @Nonnull String userId,
-                                  @Nonnull Map<String, ?> filteredAttributes,
-                                  @Nonnull ProjectConfig projectConfig) {
-        return getVariation(experiment, userId, filteredAttributes, projectConfig, Collections.emptyList(), DefaultDecisionReasons.newInstance());
+    public DecisionResponse<Variation>  getVariation(@Nonnull Experiment experiment,
```
thomaszurkan-optimizely, line 184: These methods are all marked as nullable return value.  However, the usage does not test for null:

```
Variation variation = decisionService.getVariation(experiment, userId, copiedAttributes, projectConfig).getResult();

File: core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java [hunk 4]
```diff
@@ -98,27 +94,26 @@ public Boolean evaluate(ProjectConfig config,
...
                 }
             }
         } catch (UnknownMatchTypeException | UnexpectedValueTypeException e) {
-            String message = reasons.addInfo("Audience condition \"%s\" " + e.getMessage(), this);
-            logger.warn(message);
+            logger.warn("Audience condition \"{}\" " + e.getMessage(), this);
         } catch (NullPointerException e) {
-            String message = reasons.addInfo("attribute or value null for match %s", match != null ? match : "legacy condition");
-            logger.error(message, e);
+            logger.error("attribute or value null for match {}", match != null ? match : "legacy condition", e);
```
thomaszurkan-optimizely, line 116: if the only reason to have reasons is for debugging, then, not having these doesn't make sense.  also, we need to make sure we are doing the same for other sdks.  We were including attribute failures before.

thomaszurkan-optimizely (changes requested):
Looks much cleaner.  There were a few comments/corrections.  Did we downscope reasons with this change?

core-api/src/main/java/com/optimizely/ab/Optimizely.java
jaeopt: A good catch. I'll change them all @Nonnull.

core-api/src/main/java/com/optimizely/ab/bucketing/DecisionService.java
jaeopt: I'll change them all @Nonnull.

core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
jaeopt: I agree it's a good to have in reasons, so was added before. Found that this is the only meaningful message from conditions. It's quite expensive to apply the unused reasons to all Conditions, that's why I decided to drop this from reasons. We provide other relevant reasons like "does not meet audience condition", ... and they can take a look at log messages for investigation if needed.

thomaszurkan-optimizely (approved):
LGTM

PR #417, opened 2021-01-08
feat(decide): clone user-context before calling optimizely decide

jaeopt:
## Summary
Clone the current user-context and pass it to optimizely.decide()
- user context can be changed while processing the decide request.
- with this fix, decide can return the copy of the original user-context as a part of the decision

## Test plan
- Validate user-context are copied and passed as expected

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/OptimizelyUserContext.java
+++ b/core-api/src/main/java/com/optimizely/ab/OptimizelyUserContext.java
@@ -61,11 +61,15 @@ public class OptimizelyUserContext {
     }
 
     public Map<String, Object> getAttributes() {
-        return new HashMap<String, Object>(attributes);
+        return attributes;
     }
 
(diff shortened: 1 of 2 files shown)


core-api/src/main/java/com/optimizely/ab/OptimizelyUserContext.java
```diff
@@ -61,13 +61,17 @@ public String getUserId() {
...
     public Map<String, Object> getAttributes() {
-        return new HashMap<String, Object>(attributes);
+        return attributes;
     }
 
     public Optimizely getOptimizely() {
         return optimizely;
     }
 
+    public OptimizelyUserContext copy() {
```
msohailhussain, line 71: What about implementing CloneableInterface

msohailhussain (approved):
lgtm

Issue #416, opened 2020-12-30
Socket timeout occurs frequently while fetching the datafile


UtsaVakani:
I am facing an issue similar to #360 (frequent socket timeouts) while fetching the data file .

Since , the timeout is hard coded as 10 secs in HttpClientUtils.java , we can't manually configure it using the SDK .

I am using 3.3.0 version of  `com.optimizely.ab:core-api` and `com.optimizely.ab:core-httpclient-impl` .

I also checked with the latest SDK version , but it also has the same issue .

Is there a fix/workaround already available for the same ?

Thanks,
Utsav

jaeopt:
@UtsaVakani we released a rev (3.8.2) with this fix. Let us know if it works for you.
Note that we switched to MavenCentral (Bintray/JCenter sunset) and you can find the new version (3.8.2) in there. 
We'll upload new versions to MavenCentral while serving old versions in JCenter until Feb. 01, 2022.

PR #419, opened 2021-01-23
fix: close the closable response.

thomaszurkan-optimizely:
## Summary
- close the closable response from apache httpclient.
- 
## Issues
- "Possible fix for #416"


jaeopt (approved):
LGTM

PR #420, opened 2021-01-27
Fix: SendImpressionEvent will return false when event is not sent

mnoman09:
## Summary
- SendImpressionEvent will return false when event is not sent and return true when sent successfully
- In case of userEvent is null it will return false

### DIFF
--- a/core-api/src/test/java/com/optimizely/ab/OptimizelyUserContextTest.java
+++ b/core-api/src/test/java/com/optimizely/ab/OptimizelyUserContextTest.java
@@ -1,6 +1,6 @@
 /**
  *
- *    Copyright 2020, Optimizely and contributors
+ *    Copyright 2021, Optimizely and contributors
  *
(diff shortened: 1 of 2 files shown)


jaeopt (approved):
LGTM - I added more tests for decisionEventDispatched

core-api/src/test/java/com/optimizely/ab/OptimizelyUserContextTest.java
```diff
@@ -495,6 +495,112 @@ public void decide_doNotSendEvent_withOption() {
         OptimizelyDecision decision = user.decide(flagKey, Arrays.asList(OptimizelyDecideOption.DISABLE_DECISION_EVENT));
 
         assertEquals(decision.getVariationKey(), "variation_with_traffic");
+
```
msohailhussain, line 498: please update Copyright header.

msohailhussain (approved):
lgtm

PR #423, opened 2021-02-15
Fix: GetVariationForFeatureInRollout to return null decisionResponse when rolloutRulesLength is zero

mnoman09:
## Summary
- GetVariationForFeatureInRollout to return null decisionResponse when rolloutRulesLength is zero

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/bucketing/DecisionService.java
+++ b/core-api/src/main/java/com/optimizely/ab/bucketing/DecisionService.java
@@ -283,6 +283,9 @@ public class DecisionService {
 
         // for all rules before the everyone else rule
         int rolloutRulesLength = rollout.getExperiments().size();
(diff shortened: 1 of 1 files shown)


jaeopt (approved):
LGTM
@mnoman09 What is the implication of this fix? Is this a bug fix that need be released immediately or just for FSC testing support?

mnoman09:
> LGTM
> @mnoman09 What is the implication of this fix? Is this a bug fix that need be released immediately or just for FSC testing support?

This is a bug fix. It was causing an exception on [line 37](https://github.com/optimizely/java-sdk/blob/45453d96d1ac7952368b0b74893171a8bfc6ae72/core-api/src/main/java/com/optimizely/ab/bucketing/DecisionService.java#L317) of array index out of bound in case of rollout length is zero.

PR #427, opened 2021-03-02
chore: change maven-central signing key to base64

jaeopt:
## Summary
- Change maven-central signing key to base64 to avoid travis special chars escape requirements


msohailhussain (approved):
lgtm

PR #431, opened 2021-03-05
fix: add configurable evictIdleConnections

jaeopt:
## Summary
- With periodic polling, datafile fetches may fail with timeout intermittently. 
- Add configurable "evictIdleConnections" to HttpProjectConfigManager to force close persistent connection after idle time (evict after 1min idle time by default).

## Issues
- OASIS-7421


msohailhussain (approved):
lgtm

PR #439, opened 2021-07-22
[DO NOT REVIEW] - Audiences using interface and serialize method

The-inside-man:
## Summary
- Audience Implementation using Interfaces
- Contains all new Config V2 elements

## Test plan
- FSC

## Issues
- "OASIS-7813"


core-api/src/main/java/com/optimizely/ab/config/audience/AndCondition.java
```diff
@@ -27,7 +27,7 @@
 /**
  * Represents an 'And' conditions condition operation.
  */
-public class AndCondition<T> implements Condition<T> {
+public class  AndCondition<T> implements Condition<T> {
```
keppel2, line 30: ...

The-inside-man: This is just a test, don’t worry about this PR please.

PR #441, opened 2021-08-18
Add support for custom CloseableHttpClient for Proxy Support

The-inside-man:
## Summary
- Adding additional newDefaultInstance method to support custom CloseableHttpClient object
- Allows for customs object to support Proxy Settings

Customer support

## Test plan
- FSC and TestCase added
## Issues
- "OASIS-7899"

### DIFF
--- a/core-httpclient-impl/src/main/java/com/optimizely/ab/OptimizelyFactory.java
+++ b/core-httpclient-impl/src/main/java/com/optimizely/ab/OptimizelyFactory.java
@@ -23,6 +23,7 @@ import com.optimizely.ab.event.BatchEventProcessor;
 import com.optimizely.ab.event.EventHandler;
 import com.optimizely.ab.internal.PropertyUtils;
 import com.optimizely.ab.notification.NotificationCenter;
+import org.apache.http.impl.client.CloseableHttpClient;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -48,6 +49,7 @@ import java.util.concurrent.TimeUnit;
  */
 public final class OptimizelyFactory {
     private static final Logger logger = LoggerFactory.getLogger(OptimizelyFactory.class);
+    private static CloseableHttpClient customHttpClient;
 
     /**
      * Convenience method for setting the maximum number of events contained within a batch.
@@ -221,7 +223,16 @@ public final class OptimizelyFactory {
             return newDefaultInstance(() -> null);
         }
 
-        return newDefaultInstance(sdkKey, null);
+        return newDefaultInstance(sdkKey, "");
+    }
+
+    public static Optimizely newDefaultInstance(String sdkKey, CloseableHttpClient httpClient) {
+        if (sdkKey == null || httpClient == null) {
+            logger.error("Must provide an sdkKey and httpClient, returning non-op Optimizely client");
+            return newDefaultInstance(() -> null);
+        }
+        customHttpClient = httpClient;
+        return newDefaultInstance(sdkKey);
     }
 
     /**
@@ -234,6 +245,7 @@ public final class OptimizelyFactory {
      * @return A new Optimizely instance
      */
     public static Optimizely newDefaultInstance(String sdkKey, String fallback) {
+        if (fallback.equals("")) { fallback = null; }
         String datafileAccessToken = PropertyUtils.get(HttpProjectConfigManager.CONFIG_DATAFILE_AUTH_TOKEN);
         return newDefaultInstance(sdkKey, fallback, datafileAccessToken);
     }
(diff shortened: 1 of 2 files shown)


core-httpclient-impl/src/main/java/com/optimizely/ab/OptimizelyFactory.java
```diff
@@ -221,7 +230,16 @@ public static Optimizely newDefaultInstance(String sdkKey) {
             return newDefaultInstance(() -> null);
         }
 
-        return newDefaultInstance(sdkKey, null);
+        return newDefaultInstance(sdkKey, "");
+    }
+
+    public static Optimizely newDefaultInstance(String sdkKey, CloseableHttpClient httpClient) {
```
msohailhussain, line 236: I don't think so we need this method.

core-httpclient-impl/src/main/java/com/optimizely/ab/OptimizelyFactory.java
```diff
@@ -248,12 +267,21 @@ public static Optimizely newDefaultInstance(String sdkKey, String fallback) {
...
     public static Optimizely newDefaultInstance(String sdkKey, String fallback, String datafileAccessToken) {
         NotificationCenter notificationCenter = new NotificationCenter();
-
-        HttpProjectConfigManager.Builder builder = HttpProjectConfigManager.builder()
-            .withDatafile(fallback)
-            .withNotificationCenter(notificationCenter)
-            .withSdkKey(sdkKey);
-
+        HttpProjectConfigManager.Builder builder;
+        if (customHttpClient != null) {
```
msohailhussain, line 271: this condition should only set 
`.withOptimizelyHttpClient` like we have for `datafileAccessToken`. 
or without if condition you may set null using ternary operator that won't have any impact on `OptimizelyHttpClient`. 

`.withOptimizelyHttpClient( customHttpClient  == null ? null : OHC instance`

msohailhussain (approved):
have few suggestions otherwise looks good to me.

core-httpclient-impl/src/main/java/com/optimizely/ab/OptimizelyFactory.java
```diff
@@ -48,6 +49,14 @@
...
 public final class OptimizelyFactory {
     private static final Logger logger = LoggerFactory.getLogger(OptimizelyFactory.class);
+    private static CloseableHttpClient customHttpClient;
+
+    public static void setCustomHttpClient(CloseableHttpClient httpClient) {
+        if (httpClient == null) {
+            logger.error("CloseableHttpClient must not be null. Please initialize object.");
+        }
+        customHttpClient = httpClient;
+    }
```
jaeopt, line 59: I think we can remove this method (not consistent with the rest of set methods here for property settings) and feed into newInstance parameter.

core-httpclient-impl/src/main/java/com/optimizely/ab/OptimizelyFactory.java
jaeopt: What about removing this and add httpClient option to "newDefaultInstance(String sdkKey, String fallback, String datafileAccessToken, ClosableHttpClient httpClient)? Then no need for extra set method. Considering demands for customized httpclient, it may be reasonable.

jaeopt (changes requested):
Looks good. A couple of suggestions.

core-httpclient-impl/src/main/java/com/optimizely/ab/OptimizelyFactory.java
msohailhussain: we can add default null in newDefaultInstance but that wouldn't be easy for the customer to initialize using your suggested method.

core-httpclient-impl/src/main/java/com/optimizely/ab/OptimizelyFactory.java
```diff
@@ -48,6 +49,7 @@
  */
 public final class OptimizelyFactory {
     private static final Logger logger = LoggerFactory.getLogger(OptimizelyFactory.class);
+    private static CloseableHttpClient customHttpClient;
```
jaeopt, line 52: This should be removed

core-httpclient-impl/src/main/java/com/optimizely/ab/OptimizelyFactory.java
```diff
@@ -244,14 +246,19 @@ public static Optimizely newDefaultInstance(String sdkKey, String fallback) {
      * @param sdkKey   SDK key used to build the ProjectConfigManager.
      * @param fallback Fallback datafile string used by the ProjectConfigManager to be immediately available.
      * @param datafileAccessToken  Token for authenticated datafile access.
+     * @param customHttpClient  Customizable CloseableHttpClient to build OptimizelyHttpClient.
      * @return A new Optimizely instance
      */
-    public static Optimizely newDefaultInstance(String sdkKey, String fallback, String datafileAccessToken) {
+    public static Optimizely newDefaultInstance(String sdkKey, String fallback, String datafileAccessToken, CloseableHttpClient customHttpClient) {
```
jaeopt, line 252: Can we add this new method in addition to existing (sdkKey, fallback, datafileAccessToken)? This change may break for existing clients.

jaeopt (changes requested):
Changes look good. A couple of more suggestions.

PR #443, opened 2021-08-23
Update conditions to use StringBuilder instead of StringJoiner 

The-inside-man:
## Summary
- Changing StringJoiner to StringBuilder for compatibility with Android

Android does not support StringJoiner at this time

## Test plan
FSC and Unit Tests locally

## Issues
- N/A


jaeopt (approved):
LGTM

PR #444, opened 2021-08-25
Create experimentRules array directly instead of from Map Values

The-inside-man:
## Summary
- Build ExperimentRules array instead of from Map.values()

Causing an order issue with Android-SDK

## Test plan
FSC

## Issues
- N/A

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/optimizelyconfig/OptimizelyConfigService.java
+++ b/core-api/src/main/java/com/optimizely/ab/optimizelyconfig/OptimizelyConfigService.java
@@ -26,6 +26,7 @@ public class OptimizelyConfigService {
     private ProjectConfig projectConfig;
     private OptimizelyConfig optimizelyConfig;
     private List<OptimizelyAudience> audiences;
+    private List<OptimizelyExperiment> experimentRules;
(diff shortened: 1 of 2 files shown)


core-api/src/main/java/com/optimizely/ab/optimizelyconfig/OptimizelyConfigService.java
```diff
@@ -267,17 +266,23 @@ String getExperimentFeatureKey(String experimentId) {
...
+    Map<String, OptimizelyExperiment> getExperimentsMapForFeature(List<String> experimentIds) {
         if (experimentIds == null) {
             return Collections.emptyMap();
         }
 
+        List<OptimizelyExperiment> experimentRulesList = new ArrayList<>();
+
         Map<String, OptimizelyExperiment> optimizelyExperimentKeyMap = new HashMap<>();
         for (String experimentId : experimentIds) {
             Experiment experiment = projectConfig.getExperimentIdMapping().get(experimentId);
```
jaeopt, line 278: I think this line can be removed.

jaeopt (changes requested):
Looks good. A clean up suggested.

PR #447, opened 2021-09-20
chore: clean up datafile version log message

jaeopt:
## Summary
- When datafile is parsed first time, it returns "null" for the previous version. The log message with "null previous version" can be confusing. Clean it up.

## Issues
- "OASIS-7388"


The-inside-man (approved):
LGTM!

PR #451, opened 2021-10-14
feat(ForcedDecisions): add forced-decisions APIs to OptimizelyUserContext

The-inside-man:
## Summary
Add a set of new APIs for forced-decisions to OptimizelyUserContext:
- setForcedDecision
- getForcedDecision
- removeForcedDecision
- removeAllForcedDecisions

## Test plan
- unit tests for the new APIs
- FSC tests with new test cases


core-api/src/test/java/com/optimizely/ab/OptimizelyUserContextTest.java
```diff
@@ -1445,12 +1533,39 @@ public void findValidatedForcedDecisionWithoutRuleKey() {
...
+        assertEquals(variationKey, optimizelyUserContext.getForcedDecision(optimizelyDecisionContext).getVariationKey());
+
+        // Test to confirm decide uses proper FD
+        OptimizelyDecision decision = optimizelyUserContext.decide(flagKey, Arrays.asList(OptimizelyDecideOption.INCLUDE_REASONS));
+
+        assertNotNull(decision);
+        assertTrue(decision.getReasons().contains(
+            String.format("Variation (%s) is mapped to flag (%s), rule (%s) and user (%s) in the forced decision map.", variationKey, flagKey, ruleKey, userId)
+        ));
+    }
```
jaeopt, line 1567: Can we make this test separately for each of F->D, E->D, and D->D?
And for each case, validate impression event and notification fired with the correct contents.

jaeopt (changes requested):
All the changes look great!
I suggested a couple of more changes and adding tests for events/notifications.

core-api/src/main/java/com/optimizely/ab/OptimizelyDecisionContext.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public String getRuleKey() {
+        return ruleKey != null ? ruleKey : OPTI_NULL_RULE_KEY;
+    }
+
+    public String getKey() {
+        StringBuilder keyBuilder = new StringBuilder();
+        keyBuilder.append(flagKey);
+        keyBuilder.append(getRuleKey());
+        return keyBuilder.toString();
```
jaeopt, line 45: Let's add "-$opt$-" between flagKey and ruleKey to avoid accidental conflicts like
{flagKey, ruleKey} = {A, B}
{flagKey, ruleKey} = {AB, ""}
This can be a reference for other SDKs using the flatmap.

jaeopt (changes requested):
One more request

core-api/src/main/java/com/optimizely/ab/OptimizelyDecisionContext.java
The-inside-man: Good point! Added!

core-api/src/main/java/com/optimizely/ab/OptimizelyUserContext.java
```diff
@@ -179,113 +190,69 @@ public void trackEvent(@Nonnull String eventName) throws UnknownEventTypeExcepti
...
-            Map<String, String> existingForcedDecisions = forcedDecisionsMap.get(flagKey);
-            existingForcedDecisions.put(ruleKey, variationKey);
-        } else {
-            Map<String, String> forcedDecision = new HashMap<>();
-            forcedDecision.put(ruleKey, variationKey);
-            forcedDecisionsMap.put(flagKey, forcedDecision);
+        // Check if the forcedDecisionsMap has been initialized yet or not
+        if (forcedDecisionsMap == null ){
+            // Thread-safe implementation of HashMap
+            forcedDecisionsMap = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
The-inside-man: Decided to use ConcurrentHashMap. Has some better performance and fixes that synchronizedMap doesnt.

core-api/src/main/java/com/optimizely/ab/OptimizelyUserContext.java
```diff
@@ -56,6 +53,20 @@ public OptimizelyUserContext(@Nonnull Optimizely optimizely,
...
+                                 @Nonnull Map<String, ?> attributes,
+                                 @Nullable Map<String, OptimizelyForcedDecision> forcedDecisionsMap) {
+        this.optimizely = optimizely;
+        this.userId = userId;
+        if (attributes != null) {
+            this.attributes = Collections.synchronizedMap(new HashMap<>(attributes));
+        } else {
+            this.attributes = Collections.synchronizedMap(new HashMap<>());
+        }
+        this.forcedDecisionsMap = forcedDecisionsMap;
```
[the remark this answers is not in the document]
The-inside-man: No problem!

core-api/src/test/java/com/optimizely/ab/OptimizelyUserContextTest.java
The-inside-man: Added 3 new tests with each and creating the impression events and verifying the contents.

core-api/src/main/java/com/optimizely/ab/OptimizelyUserContext.java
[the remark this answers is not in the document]
The-inside-man: It turns out that the synchronized map allows for null as a key which is needed for the attributes. For forcedDecisionMap, null is not a valid key, ConcurrentMap does not allow Null so this makes sense to keep as ConcurrentMap as well. I think it makes more sense to keep this as is.

jaeopt (changes requested):
Looks good! One more comment.

The-inside-man:
> Looks good! One more comment.

Roger! Committed the change!

core-api/src/main/java/com/optimizely/ab/OptimizelyUserContext.java
```diff
@@ -64,7 +64,9 @@ public OptimizelyUserContext(@Nonnull Optimizely optimizely,
         } else {
             this.attributes = Collections.synchronizedMap(new HashMap<>());
         }
-        this.forcedDecisionsMap = forcedDecisionsMap != null ? forcedDecisionsMap : new ConcurrentHashMap<>();
+    if (forcedDecisionMap != null) {
```
jaeopt, line 67: nit - indent

jaeopt (approved):
LGTM with a nit

dustin-sier (approved):
LGTM!

Issue #391, opened 2020-08-07
Support Moshi as an optional JSON parser


yogurtearl:
Optimizely supports multiple optional JSON parsers. 
https://github.com/optimizely/java-sdk/blob/12acebad6c5574fa4e57662306c4d4701ee15fbd/core-api/build.gradle#L9

Moshi is a popular JSON parser, especially on Android. 
https://github.com/square/moshi

Please add optional support for moshi.

The-inside-man:
After careful review of the following issue/request, our team has determined that in order to move forward, we would kindly ask for a submission on the following page, carefully outlining what is being requested. This allows our team to work your request into our backlog of items. This is usually a result of current work load, priority tasks, size of request, etc. We appreciate your cooperation and understanding.

Please submit here: https://experimentation.ideas.aha.io/

PR #452, opened 2021-11-10
enhancement: Added support for Moshi

The-inside-man:
## Summary
- Add Moshi json parser support to gradle build

As requested by several customers

## Test plan
- FSC
- Verify Build

## Issues
- "OASIS-8052"
- [ISSUE-391](https://github.com/optimizely/java-sdk/issues/391)


jaeopt (changes requested):
Don't we need to add ProjectConfig parser for Moshi? It must be much more expensive to add the support :)

The-inside-man:
> d ProjectConfig parser fo

Ahh yes you may be right! haha Let me look at that! thank you for pointing that out!

The-inside-man:
> Don't we need to add ProjectConfig parser for Moshi? It must be much more expensive to add the support :)

Ahh yes you may be right! haha Let me look at that! thank you for pointing that out!

The-inside-man:
Closing PR for now as will need to be worked and backlogged appropriately.

PR #453, opened 2021-11-22
Added isExperiment running check to avoid evaluating paused rollout experiments

mnoman09:
## Summary
- IsExperiment running check is added in rollout evaluation to make sure paused rollout experiment gets skipped.

## Test plan
All FSC and unit tests should pass.


jaeopt (approved):
LGTM

core-api/src/main/java/com/optimizely/ab/bucketing/DecisionService.java
```diff
@@ -307,7 +307,10 @@ public DecisionService(@Nonnull Bucketer bucketer,
 
         int index = 0;
         while (index < rolloutRulesLength) {
-
+            if (!rollout.getExperiments().get(index).isRunning()) {
```
The-inside-man, line 310: Should this check the status of the experiment to be "paused" rather than is running since there are 5 other status'? Or is this behavior supposed to be for every status other than running?

PR #454, opened 2021-12-03
fix(ForcedDecision): remove config-ready check from forced-decision apis

The-inside-man:
## Summary
- Remove config-ready check from forced-decision apis

## Test plan
- Unit and FSC tests


jaeopt (approved):
LGTM

PR #455, opened 2021-12-06
fix:update OptimizelyDecisionContext to check null for flagKey

The-inside-man:
## Summary
- Add null check to OptimizelyDecisionContext constructor for flagKey

We want to confirm the user is not passing a null value as flagKey in addition to the @NonNull annotation provided to ensure the object is valid

## Test plan
- FSC

## Issues
N/A


jaeopt (approved):
LGTM

PR #459, opened 2022-01-20
fix:Change Log level from WARN to DEBUG for number values in Audience Conditions for revisionNumber

The-inside-man:
## Summary
- Changing log level from WARN to DEBUG for Audience Conditions when UnknownMatchTypeException or UnexpectedValueTypeException are caught.

Several customers have expressed concerns with logs being blown up with warnings due to usage of numbers for revisionNumber and other instances that require the use of a number value in their application where the SDK expects a string value. To avoid warnings in production environments, the log-level has been changed to DEBUG only now for those instances.

## Test plan
- New Unit Test
- FSC

## Issues
- "OASIS-8129"

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
+++ b/core-api/src/main/java/com/optimizely/ab/config/audience/UserAttribute.java
@@ -111,7 +111,7 @@ public class UserAttribute<T> implements Condition<T> {
                 }
             }
         } catch (UnknownMatchTypeException | UnexpectedValueTypeException e) {
-            logger.warn("Audience condition \"{}\" " + e.getMessage(), this);
+            logger.debug("Audience condition \"{}\" " + e.getMessage(), this);
         } catch (NullPointerException e) {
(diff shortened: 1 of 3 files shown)


The-inside-man:
@jaeopt Could you please take a look at this. All I have done is change the Log level for when the exception is caught to avoid blowing up the logs for the unknown and incorrect types.

dustin-sier (approved):
LGTM. I checked C# (its debug) and looking through Go to see if we're doing similar.

jaeopt (changes requested):
I think we should stick to WARN level for UnknownMatchTypeException and UnexpectedValueTypeException. It must be an error on their settings.
The one with potentially exploding log messages is when they do not provide attribute for audience (instead of wrong type). It's interesting that the code already logs missing attribute as a debug level with correct messages (line 97).

jaeopt:
We'll keep the current WARNING level for these issues. It's confirmed that we currently generate a DEBUG level log for missing attribute as expected. The only  issue was with Sematic Version, which has been fixed with https://github.com/optimizely/java-sdk/pull/463.

PR #460, opened 2022-01-21
fix(NotificationManager): add thread-safety to NotificationManager

jaeopt:
## Summary
- Fix NotificationManager to be thread-safe (add-handler and send-notifications can happen concurrently)

## Test plan
- Add a unit test concurrently adding handlers and sending notifications.

## Issues
- OASIS-8135


The-inside-man (approved):
LGTM!

PR #473, opened 2022-06-08
Fixing the git action build ubuntu to 18.04 

mnoman09:
## Summary
- with ubuntu latest tests were failing without any particular reason and no fix was working so as we were testing on travis using ubuntu trusty so we just changed the ubuntu version to fix 18.04 to avoid further issues 

## Test plan
all tests should pass


msohailhussain (approved):
lgtm

PR #481, opened 2022-07-29
feat: Add ODP GraphQL Api Interface

zashraf1985:
## Summary
1. Added an ODPApiManager interface and its Default implementation that implements the graphQL call to fetch qualified segments.
2. The mechanism to get default parser is tightly coupled with Project Config module. I created a separate Factory class to get the default parser. This can be used by any future implementations that need a default parser.
3. Added parsing for all four supported parsers to parse valid and error responses.

## Test plan
1. Added new unit tests to test all the new code.
2. Existing unit tests pass
3. Existing Full stack compatibility suite tests pass

## Issues
[OASIS-8384](https://optimizely.atlassian.net/browse/OASIS-8384)


core-httpclient-impl/src/main/java/com/optimizely/ab/odp/DefaultODPApiManager.java
```diff
@@ -0,0 +1,179 @@
...
+           }
+         }
+       ],
+       "data": {
+         "customer": null
+       }
+     }
+    */
+    @Override
+    public String fetchQualifiedSegments(String apiKey, String apiEndpoint, String userKey, String userValue, List<String> segmentsToCheck) {
```
Mat001, line 133: Name "fetchQualifiedSegments" is already used in upcoming "ODPSegmentManager" (per Jae's naming). Risk of name conflict. Unless you use different names, but that might be a deviation from naming consistency across SDKs?

core-api/src/main/java/com/optimizely/ab/odp/ODPApiManager.java
```diff
@@ -0,0 +1,22 @@
...
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+package com.optimizely.ab.odp;
+
+import java.util.List;
+
+public interface ODPApiManager {
+    String fetchQualifiedSegments(String apiKey, String apiEndpoint, String userKey, String userValue, List<String> segmentsToCheck);
```
Mat001, line 21: Name already "reserved" for the method in ODPSegmentManager? This one should be fetchSegments, no? (consistency across SDKs)

core-httpclient-impl/src/test/java/com/optimizely/ab/odp/DefaultODPApiManagerTest.java
```diff
@@ -0,0 +1,116 @@
...
+    @Test
+    public void returnNullWhenStatusIsNot200AndLogError() throws Exception {
+        setupHttpClient(500);
+        ODPApiManager apiManager = new DefaultODPApiManager(mockHttpClient);
+        String responseString = apiManager.fetchQualifiedSegments("key", "endPoint", "fs_user_id", "test_user", Arrays.asList("segment_1", "segment_2"));
+        verify(mockHttpClient, times(1)).execute(any(HttpPost.class));
+        logbackVerifier.expectMessage(Level.ERROR, "Unexpected response from ODP server, Response code: 500, null");
+        assertEquals(null, responseString);
+    }
+}
```
Mat001, line 116: What about the test for network issues, such as internet is down, where status code is not returned. Request is not reaching the server and bounces off?

core-api/src/main/java/com/optimizely/ab/odp/ODPApiManager.java
Mat001: Not 100% sure, maybe I'm wrong... you have ODPAPIManager interface that defines fetchQualifiedSegments method, but there is also method fetchSegments in Swift PR that is located in the GraphQL API Manager. Not sure how you are overriding, but it seems smth not quite right...unless it's my weak java knowledge

Mat001 (changes requested):

core-api/src/main/java/com/optimizely/ab/odp/ODPApiManager.java
zashraf1985: Design of java is a little different than other sdks. For all the external calls, we need to create an interface and pass it in to the core SDK from the outside. That is why i created ODPAPIManager which is an interface that will contain all the external calls such as graphQL and ODPEvents. Then i am providing a DefaultODPApiManager as a default implementation. user will use this by default but can pass in their own implementation. To me `fethQualifiedSegments` makes more sense because thats exactly what this call does. It fetches the qualified segments not all or just any segments.

jaeopt, line 21: It looks like you plan to open all ODP APIs here to public for customization support. It looks good for now and we can continue discussion the final apis.

jaeopt (changes requested):
A refactoring for sharing multiple JSON parsers suggested if it's not too much.
I need to see the next stage to see how you want to implement all. It looks good as is!

core-api/src/main/java/com/optimizely/ab/internal/JsonParserProvider.java
```diff
@@ -0,0 +1,74 @@
...
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+package com.optimizely.ab.internal;
+
+import com.optimizely.ab.config.parser.MissingJsonParserException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public enum JsonParserProvider {
```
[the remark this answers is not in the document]
zashraf1985: I wanted to reuse code from there but it was tightly couples. I wanted to refactor DefaultConfigParser to use this one but i thought i will add a comment and do that later as a separate ticket. Right now it will become too much work. what do you think?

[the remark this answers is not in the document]
zashraf1985: The idea is to make DefaultConfigParser and whatever else needs a parser use this code eventually. i am just too afraid to touch DafaultConfigParser in this PR

core-api/src/main/java/com/optimizely/ab/odp/ODPApiManager.java
zashraf1985: Yes, this will contain all the methods that make an external call. ODPEvents calls will also go here

core-api/src/main/java/com/optimizely/ab/internal/JsonParserProvider.java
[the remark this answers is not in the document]
jaeopt: I leave it to you.

core-httpclient-impl/src/main/java/com/optimizely/ab/odp/DefaultODPApiManager.java
```diff
@@ -0,0 +1,179 @@
...
+        HttpPost request = new HttpPost(apiEndpoint);
```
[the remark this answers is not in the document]
zashraf1985: I see this as the lowest level method that will make the actual call that is why it's getting the full endPoint. Users might provide their own implementation so it made more sense for it to expect the full endPoint. My plan is to build this endpoint using the host and path in the `ODPSegmentManager` and pass it to this one. What do you think?

core-api/src/main/java/com/optimizely/ab/internal/JsonParserProvider.java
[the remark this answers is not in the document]
zashraf1985: Cool! I will add a comment in the DefaultConfigParser mentioning the same. I will also add a ticket to refactor this and put it in the tech debt backlog.

core-httpclient-impl/src/main/java/com/optimizely/ab/odp/DefaultODPApiManager.java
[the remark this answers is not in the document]
jaeopt: That's what I thought. If we plan to support custom endpoints, this makes sense. Let's move on and review again at the top level. 
One thing we also should consider is that if clients want to implement their own segment services (replacing all our default ODPManager), they can determine segments using their own solution and directly feed the "qualifedSegments" into OptimizelyUserContext, instead of injecting own ODPManager into our SDKs.

jaeopt (approved):
LGTM

Mat001 (approved):
lgtm

PR #482, opened 2022-08-03
feat: Added a generic LRUCache interface and a default implementation

zashraf1985:
## Summary
Added an interface for a generic LRUCache and a default Implementation.

## Test plan
- All existing tests pass.
- Added unit tests for the default cache implementation.

## Issues
[OASIS-8385](https://optimizely.atlassian.net/browse/OASIS-8385)

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/internal/DefaultLRUCache.java
+++ b/core-api/src/main/java/com/optimizely/ab/internal/DefaultLRUCache.java
@@ -112,6 +112,19 @@ public class DefaultLRUCache<T> implements LRUCache<T> {
         }
     }
 
+    public T peek(String key) {
+        if (maxSize == 0) {
+            // Cache is disabled when maxSize = 0
+            return null;
+        }
+
+        if (hashMap.containsKey(key)) {
+            return hashMap.get(key).value;
+        }
+
+        return null;
+    }
+
     public void reset() {
         synchronized (lock) {
             linkedList.clear();
--- a/core-api/src/main/java/com/optimizely/ab/internal/DefaultLRUCache.java
+++ b/core-api/src/main/java/com/optimizely/ab/internal/DefaultLRUCache.java
@@ -32,6 +32,7 @@ public class DefaultLRUCache<T> implements Cache<T> {
     private Integer maxSize;
 
     private Long timeoutMillis;
+    
     @VisibleForTesting
     final LinkedHashMap<String, ItemWrapper> linkedHashMap = new LinkedHashMap<String, ItemWrapper>(16, 0.75f, true) {
         @Override
--- a/core-api/src/main/java/com/optimizely/ab/internal/DefaultLRUCache.java
+++ b/core-api/src/main/java/com/optimizely/ab/internal/DefaultLRUCache.java
@@ -1,0 +1,134 @@
+/**
+ *
+ *    Copyright 2022, Optimizely
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+package com.optimizely.ab.internal;
+
+import com.optimizely.ab.annotations.VisibleForTesting;
+
+import java.util.*;
+
+public class DefaultLRUCache<T> implements LRUCache<T> {
+
+    private final Object lock = new Object();
+
+    private Integer maxSize;
+
+    private Long timeoutMillis;
+
+    final LinkedList<ItemWrapper> linkedList = new LinkedList<>();
+
+    final Map<String, ItemWrapper> hashMap = new HashMap<>();
+
+    public DefaultLRUCache() {
+        this.maxSize = DEFAULT_MAX_SIZE;
+        this.timeoutMillis = (long) (DEFAULT_TIMEOUT_SECONDS * 1000);
+    }
+
+    public void setMaxSize(Integer size) {
+        Integer sizeToSet = size;
+        if (size < 0 ) {
+            sizeToSet = DEFAULT_MAX_SIZE;
+        }
+
+        synchronized (lock) {
+            if (linkedList.size() > sizeToSet) {
+                // If cache is bigger than the new maxSize, remove additional items from the end.
+                int itemsToTrim = linkedList.size() - sizeToSet;
+                for (int i = 0; i < itemsToTrim; i++) {
+                    ItemWrapper item = linkedList.removeLast();
+                    hashMap.remove(item.key);
+                }
+            }
+            this.maxSize = sizeToSet;
+        }
+    }
+
+    public void setTimeout(Long timeoutSeconds) {
+        this.timeoutMillis = timeoutSeconds * 1000;
+    }
+
+    public void save(String key, T value) {
+        if (maxSize == 0) {
+            // Cache is disabled when maxSize = 0
+            return;
+        }
+        synchronized (lock) {
+            ItemWrapper item = new ItemWrapper(key, value);
+
+            if (hashMap.containsKey(key)) {
+                ItemWrapper oldItem = hashMap.get(key);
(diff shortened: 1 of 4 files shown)


core-api/src/main/java/com/optimizely/ab/internal/DefaultLRUCache.java
```diff
@@ -0,0 +1,147 @@
...
+
+    public DefaultLRUCache() {
+        this.maxSize = DEFAULT_MAX_SIZE;
+        this.timeoutMillis = (long) (DEFAULT_TIMEOUT_SECONDS * 1000);
+    }
+
+    public void setMaxSize(Integer size) {
+        Integer sizeToSet = size;
+        if (size < 0 ) {
+            sizeToSet = DEFAULT_MAX_SIZE;
```
jaeopt, line 43: either way is fine, but let's use "sizeToSet = 0" for all SDKs.

core-api/src/main/java/com/optimizely/ab/internal/DefaultLRUCache.java
```diff
@@ -0,0 +1,147 @@
...
+
+        synchronized (lock) {
+            if (linkedList.size() > sizeToSet) {
+                // If cache is bigger than the new maxSize, remove additional items from the end.
+                int itemsToTrim = linkedList.size() - sizeToSet;
+                for (int i = 0; i < itemsToTrim; i++) {
+                    ItemWrapper item = linkedList.removeLast();
+                    hashMap.remove(item.key);
+                }
+            }
```
jaeopt, line 54: Do we need this support for dynamic cache size changes? We can allow them cache size setting only when app starts.

core-api/src/main/java/com/optimizely/ab/internal/DefaultLRUCache.java
```diff
@@ -0,0 +1,147 @@
...
+        if (maxSize == 0) {
+            // Cache is disabled when maxSize = 0
+            return;
+        }
+        synchronized (lock) {
+            ItemWrapper item = new ItemWrapper(key, value);
+
+            if (hashMap.containsKey(key)) {
+                ItemWrapper oldItem = hashMap.get(key);
+                linkedList.remove(oldItem);
```
jaeopt, line 73: Isn't this linear search? It may be a factor in large cache.

core-api/src/test/java/com/optimizely/ab/internal/DefaultLRUCacheTest.java
```diff
@@ -0,0 +1,171 @@
...
+    @Test
+    public void createSaveAndLookupOneItem() {
+        LRUCache<String> cache = new DefaultLRUCache<>();
+        assertNull(cache.lookup("key1"));
+        cache.save("key1", "value1");
+        assertEquals("value1", cache.lookup("key1"));
+    }
+
+    @Test
+    public void saveAndLookupMultipleItems() {
```
jaeopt, line 37: This test looks good. It will be good to have the same ordering switch tests for save ops as well.

core-api/src/main/java/com/optimizely/ab/internal/DefaultLRUCache.java
```diff
@@ -0,0 +1,147 @@
...
+                linkedList.addFirst(item);
+                hashMap.replace(key, item);
+                return;
+            }
+
+            if (linkedList.size() == maxSize) {
```
jaeopt, line 79:
```suggestion
            if (linkedList.size() >= maxSize) {
```

core-api/src/test/java/com/optimizely/ab/internal/DefaultLRUCacheTest.java
```diff
@@ -0,0 +1,171 @@
...
+        assertEquals(2, cache.linkedList.size());
+        assertEquals(2, cache.hashMap.size());
+
+        assertEquals(Arrays.asList("segment5", "segment6"), cache.lookup("user3"));
+        assertEquals(Arrays.asList("segment3", "segment4"), cache.lookup("user2"));
+        assertNull(cache.lookup("user1"));
+    }
+
+    @Test
+    public void whenMaxSizeIsReducedInBetween() {
```
jaeopt, line 115: We can remove this test

core-api/src/test/java/com/optimizely/ab/internal/DefaultLRUCacheTest.java
```diff
@@ -0,0 +1,171 @@
...
+        assertEquals(Arrays.asList("segment5", "segment6"), cache.lookup("user3"));
+        assertNull(cache.lookup("user1"));
+        assertNull(cache.lookup("user2"));
+
+        assertEquals(1, cache.linkedList.size());
+        assertEquals(1, cache.hashMap.size());
+    }
+
+    @Test
+    public void saveAndPeekItems() {
```
jaeopt, line 139: A test for reset()?

jaeopt (changes requested):
Looks good! A few suggestions and clarifications.

core-api/src/main/java/com/optimizely/ab/internal/DefaultLRUCache.java
zashraf1985: Done! Defaulting to zero now when max size is negative

core-api/src/main/java/com/optimizely/ab/internal/DefaultLRUCache.java
zashraf1985: I changed it to only allow when maxSize passed in is less than the current size of the cache.

core-api/src/main/java/com/optimizely/ab/internal/DefaultLRUCache.java
zashraf1985: This is a good catch. Because i was passing the item itself and it uses a doubly linked list, i thought the item should have references to its neighbors and should be able to cleanly remove itself in O(1) but it looks like the actual node with references is managed internally and `remove` actually searched for the item linearly. I then went ahead and redid the whole implementation using `LunkedHashMap`

core-api/src/test/java/com/optimizely/ab/internal/DefaultLRUCacheTest.java
zashraf1985: Done!

core-api/src/test/java/com/optimizely/ab/internal/DefaultLRUCacheTest.java
zashraf1985: Modified the test to check the changed implementation

core-api/src/main/java/com/optimizely/ab/internal/DefaultLRUCache.java
zashraf1985: Changed the whole implementation. It does not apply anymore.

core-api/src/test/java/com/optimizely/ab/internal/DefaultLRUCacheTest.java
zashraf1985: Done!

core-api/src/main/java/com/optimizely/ab/internal/DefaultLRUCache.java
```diff
@@ -65,24 +71,9 @@ public void save(String key, T value) {
...
-            if (linkedList.size() == maxSize) {
-                ItemWrapper removedItem = linkedList.removeLast();
-                hashMap.remove(removedItem.key);
-            }
-
-            linkedList.addFirst(item);
-            hashMap.put(key, item);
+        synchronized (lock) {
+            linkedHashMap.put(key, new ItemWrapper(value));
         }
```
jaeopt, line 77: Wow, a single-line solution :)

core-api/src/main/java/com/optimizely/ab/internal/DefaultLRUCache.java
```diff
@@ -93,53 +84,32 @@ public T lookup(String key) {
...
-                // If item is stale, remove it.
-                linkedList.remove(item);
-                hashMap.remove(item.key);
+                linkedHashMap.remove(key);
             }
             return null;
         }
     }
 
-    public T peek(String key) {
```
jaeopt: It's surprising that they do not have this method in LinkedHashMap.  It should be ok without this since it's mostly for testing/debugging.

jaeopt (changes requested):
The new solution with LinkedHashMap looks great!
I suggest to remove support for dynamic cache config changes (setMaxsize/setTimeout) for simplicity and also consistency with other SDKs. Other than that, all changes look good.

jaeopt (approved):
LGTM

PR #485, opened 2022-08-18
feat: Added ODP RestApi interface for sending Events to ODP

zashraf1985:
## Summary
This module provides an internal service for ODP event REST api access.

## Test plan
- Manually tested thoroughly.
- Added new unit tests.
- All Existing Unit tests pass.
- All existing Full stack compatibility tests pass.

## Jira
[OASIS-8387](https://optimizely.atlassian.net/browse/OASIS-8387)


core-httpclient-impl/src/main/java/com/optimizely/ab/odp/DefaultODPApiManager.java
```diff
@@ -167,6 +182,43 @@ public String fetchQualifiedSegments(String apiKey, String apiEndpoint, String u
...
+
+        CloseableHttpResponse response = null;
+        try {
+            response = httpClient.execute(request);
+        } catch (IOException e) {
+            logger.error("Error retrieving response from event request", e);
+            return;
+        }
+
+        if (response.getStatusLine().getStatusCode() >= 400) {
```
jaeopt, line 212: Failures with recoverable errors (status=5xx, connection errors, etc) should be differentiated so that OdpEventManager can retry them later.

core-api/src/main/java/com/optimizely/ab/odp/ODPEvent.java
```diff
@@ -0,0 +1,64 @@
...
+
+import java.util.Map;
+
+public class ODPEvent {
+    private String type;
+    private String action;
+    private Map<String, String > identifiers;
+    private Map<String, String> data;
+
+    public ODPEvent(String type, String action, Map<String, String> identifiers, Map<String, String> data) {
```
jaeopt, line 26:
```suggestion
    public ODPEvent(String type, String action, Map<String, String> identifiers, Map<String, Object> data) {
```
It can support any type of data (String, Number, Boolean).

core-api/src/main/java/com/optimizely/ab/odp/ODPApiManager.java
```diff
@@ -19,4 +19,6 @@
 
 public interface ODPApiManager {
     String fetchQualifiedSegments(String apiKey, String apiEndpoint, String userKey, String userValue, List<String> segmentsToCheck);
+
+    void sendEvents(String apiKey, String apiEndpoint, List<ODPEvent> events);
 }
```
jaeopt, line 24: It looks like you plan to open these to public (and Android-SDK) to customize. If so, ODPEvent should be open too, but this is not desired. Let's discuss more.

jaeopt (changes requested):
Looks good. A couple of changes suggested.

core-api/src/main/java/com/optimizely/ab/odp/serializer/ODPJsonSerializerFactory.java
```diff
@@ -0,0 +1,49 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(ODPJsonSerializerFactory.class);
+
+    public static ODPJsonSerializer getSerializer() {
+        JsonParserProvider parserProvider = JsonParserProvider.getDefaultParser();
+        ODPJsonSerializer jsonSerializer = null;
+        switch (parserProvider) {
+            case GSON_CONFIG_PARSER:
+                jsonSerializer = new GsonSerializer();
+                break;
+            case JACKSON_CONFIG_PARSER:
```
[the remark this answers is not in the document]
zashraf1985: Its using the same JsonParserProvider to find out the default one. Its just exporting the implementation class separately. This is separate because it is a seralizer. I wanted to keep parser and serializers separate to avoid confusion. any more parsing or serialization done for ODP will go in to these same classes as added methods. Let me know what you think

core-api/src/main/java/com/optimizely/ab/odp/ODPEvent.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+package com.optimizely.ab.odp;
+
+import java.util.Map;
+
+public class ODPEvent {
+    private String type;
+    private String action;
+    private Map<String, String > identifiers;
+    private Map<String, String> data;
```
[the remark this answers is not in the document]
zashraf1985: Done!

core-api/src/main/java/com/optimizely/ab/odp/ODPEvent.java
```diff
@@ -0,0 +1,64 @@
...
+
+    public Map<String, String> getIdentifiers() {
+        return identifiers;
+    }
+
+    public void setIdentifiers(Map<String, String> identifiers) {
+        this.identifiers = identifiers;
+    }
+
+    public Map<String, String> getData() {
```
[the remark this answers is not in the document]
zashraf1985: Done

core-api/src/main/java/com/optimizely/ab/odp/ODPEvent.java
```diff
@@ -0,0 +1,64 @@
...
+        return data;
+    }
+
+    public void setData(Map<String, String> data) {
```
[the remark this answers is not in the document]
zashraf1985: Done

core-api/src/test/java/com/optimizely/ab/odp/serializer/ODPJsonSerializerFactoryTest.java
```diff
@@ -0,0 +1,50 @@
...
+    public void getGsonSerializerWhenNoDefaultIsSet() {
+        assertEquals(GsonSerializer.class, ODPJsonSerializerFactory.getSerializer().getClass());
+    }
+
+    @Test
+    public void getCorrectSerializerWhenValidDefaultIsProvided() {
+        PropertyUtils.set("default_parser", "JSON_CONFIG_PARSER");
+        assertEquals(JsonSerializer.class, ODPJsonSerializerFactory.getSerializer().getClass());
+    }
+
```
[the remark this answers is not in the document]
zashraf1985: Done

core-api/src/main/java/com/optimizely/ab/odp/ODPEvent.java
zashraf1985: Done

core-httpclient-impl/src/main/java/com/optimizely/ab/odp/DefaultODPApiManager.java
zashraf1985: Now returning status code from the method so that consuming code can use it and retry when needed.

core-api/src/main/java/com/optimizely/ab/odp/ODPApiManager.java
jaeopt: I fully understand opening is required to support Android and other clients. My question was about opening "ODPEvent" type. We can consider to accept serialized String or map instead here, that's how we avoid internal type.

core-api/src/main/java/com/optimizely/ab/odp/serializer/ODPJsonSerializerFactory.java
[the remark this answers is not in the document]
jaeopt: Separating parser and serializer looks good. I was talking about the future refactoring discussed already to reuse them in {datafile parser, event serializer, segments parser, and odp event serializer}.

core-api/src/test/java/com/optimizely/ab/odp/serializer/ODPJsonSerializerTest.java
```diff
@@ -70,9 +70,14 @@ private static ODPEvent createTestEvent(String index) {
...
         identifiers.put("vuid-" + index + "-3", "fs-" + index + "-3");
 
-        Map<String, String> data = new HashMap<>();
+        Map<String, Object> data = new HashMap<>();
         data.put("source", "java-sdk");
         data.put("data-1", "data-value-" + index);
+        data.put("data-num", Integer.parseInt(index));
+        data.put("data-float", 2.33);
+        data.put("data-bool-true", true);
+        data.put("data-bool-false", false);
```
jaeopt, line 79: Let's add <null> to this test, which is also supported.

core-httpclient-impl/src/main/java/com/optimizely/ab/odp/DefaultODPApiManager.java
```diff
@@ -206,17 +205,19 @@ public void sendEvents(String apiKey, String apiEndpoint, List<ODPEvent> events)
             response = httpClient.execute(request);
         } catch (IOException e) {
             logger.error("Error retrieving response from event request", e);
-            return;
+            return null;
```
jaeopt, line 208: Should we retry on "null" response?  How can we differentiate connection error (retry) from other permanent errors?

jaeopt (changes requested):
A couple of more suggestions.

core-api/src/test/java/com/optimizely/ab/odp/serializer/ODPJsonSerializerTest.java
zashraf1985: I did not get this suggestion. Can you elaborate a bit?

core-api/src/main/java/com/optimizely/ab/odp/ODPApiManager.java
zashraf1985: I would emphasize on keeping the ODPEvent type exposed here. It's only and Entity class which does not contain any business logic. Adding a string input here will increase the burden of unnecessary serialization on us. Keeping it a list of ODPEvent is easy to use and less error prone. What do you suggest?

core-api/src/test/java/com/optimizely/ab/odp/serializer/ODPJsonSerializerTest.java
jaeopt: "Let's add <nulll> to this test". I missed the key word <null> in my comment :)

core-api/src/main/java/com/optimizely/ab/odp/ODPApiManager.java
jaeopt: @zashraf1985 I understand that it'll look nice to pass the structure, but opening ODPEvent may be expensive. Following our rule, we should rename it to "OptimizelyOdpEvent" like other public type, which also should be documented. Other SDKs also should be changed for consistency. A question is that it's worth to open it for the cost? Clients do not care about the contents, just serialized and dispatched.

core-httpclient-impl/src/main/java/com/optimizely/ab/odp/DefaultODPApiManager.java
```diff
@@ -205,7 +205,7 @@ public Integer sendEvents(String apiKey, String apiEndpoint, List<ODPEvent> even
             response = httpClient.execute(request);
         } catch (IOException e) {
             logger.error("Error retrieving response from event request", e);
-            return null;
+            return 0;
```
jaeopt, line 208: Either null or 0 is fine with me. My question was that how we can tell if this failure should be retried or not at the upper level. Do you plan to retry all status of 0 and 50x?

jaeopt (changes requested):
See my other clarifications in the previous reviews.

core-httpclient-impl/src/main/java/com/optimizely/ab/odp/DefaultODPApiManager.java
zashraf1985: yes! I will document it on the method too.
1. null means there was a non recoverable error and no retry is needed.
2. 0 means an unexpected error and you should retry
3. Any other status can be handled as needed but my idea is to return on 500+

core-api/src/main/java/com/optimizely/ab/odp/ODPApiManager.java
zashraf1985: I think you are right. The only responsibility of the `httpclient` should be to take the payload, send it and return status. The serialization should be handled in EventManager. I will change it to get the string payload. The serialization code is still useful for the ODPEventManager so i will leave it there with this PR.

core-api/src/test/java/com/optimizely/ab/odp/serializer/ODPJsonSerializerTest.java
jaeopt: Testing <null>

jaeopt: Testing "null"

jaeopt: I should not use "<null>", which is removed by github :) I meant to add a type checking for "null" type.

jaeopt (approved):
LGTM

PR #489, opened 2022-10-14
feat: Added ODPManager implementation

zashraf1985:
## Summary
Added ODPManager Implementation which does the following.
1. Initializes and provides access to ODPEventManager and ODPSegmentManager
2. Provides updated ODPConfig settings to event manager and segment manager.
3. Stops Event Manager thread when closed.

## Test plan
1. Manually tested thoroughly
2. Added unit tests.

## Issues
[FSSDK-8388](https://jira.sso.episerver.net/browse/FSSDK-8388)


core-api/src/main/java/com/optimizely/ab/odp/ODPManager.java
```diff
@@ -0,0 +1,61 @@
...
+public class ODPManager {
+    private volatile ODPConfig odpConfig;
+    private final ODPSegmentManager segmentManager;
+    private final ODPEventManager eventManager;
+
+    public ODPManager(@Nonnull ODPConfig odpConfig, @Nonnull ODPApiManager apiManager) {
+        this(odpConfig, new ODPSegmentManager(odpConfig, apiManager), new ODPEventManager(odpConfig, apiManager));
+    }
+
+    public ODPManager(@Nonnull ODPConfig odpConfig, @Nonnull ODPSegmentManager segmentManager, @Nonnull ODPEventManager eventManager) {
```
jaeopt, line 31: Customization of segmentCache size/timeout and segmentCache itself as well?

jaeopt: We also need to support for "disable" entire ODP functions.

core-api/src/main/java/com/optimizely/ab/odp/ODPManager.java
```diff
@@ -0,0 +1,61 @@
...
+            segmentManager.updateSettings(odpConfig);
+            return true;
+        }
+        return false;
+    }
+
+    public void close() {
+        eventManager.stop();
+    }
+}
```
jaeopt, line 61: We do not have top-level `fetchSegments` and `sendEvent` apis. Is it a plan to open directly from `OdpEventManager` and `OdpSegmentManager`?

core-api/src/main/java/com/optimizely/ab/odp/ODPManager.java
```diff
@@ -0,0 +1,61 @@
...
+    public ODPSegmentManager getSegmentManager() {
+        return segmentManager;
+    }
+
+    public ODPEventManager getEventManager() {
+        return eventManager;
+    }
+
+    public Boolean updateSettings(String apiHost, String apiKey, List<String> allSegments) {
+        ODPConfig newConfig = new ODPConfig(apiKey, apiHost, allSegments);
```
jaeopt, line 47: Can we flush events in queue before updating apiKey/apiHost? It should not be common but when clients change the ODP account, all events in the queue should be flushed to the old account.

core-api/src/main/java/com/optimizely/ab/odp/ODPManager.java
jaeopt: It is expected to filter out api requests in this level (like "disable" discards all requests).  Also sendEvent data type should be validated here (are we doing this in OdpEventManager?)

jaeopt (changes requested):
Looks good. Some changes and clarifications suggested.

core-api/src/main/java/com/optimizely/ab/odp/ODPManager.java
zashraf1985: There are two constructors.
1. First one only takes the two mandatory arguments of `ODPConfig` and `ApiManager` and initializes everything else on default values. 
2. If you want to customize anything at all, use the second constructor which expects you to pass everything from the outside. This is where you will have a chance to initialize your own `SegmentManager` with your own `cacheOptions`.

I will recommend to keep these two options for now. We can provide other combinations at the level of factory methods to create the `SDK` instance later as needed. What do you think?

zashraf1985: I am thinking of handling disable / enable at `optimizely` level. When enabled, ODP object will be there and it will always work. When disabled, there will be no ODP instance at all. What do you suggest?

jaeopt: > There are two constructors.
> 
> 1. First one only takes the two mandatory arguments of `ODPConfig` and `ApiManager` and initializes everything else on default values.
> 2. If you want to customize anything at all, use the second constructor which expects you to pass everything from the outside. This is where you will have a chance to initialize your own `SegmentManager` with your own `cacheOptions`.
> 
> I will recommend to keep these two options for now. We can provide other combinations at the level of factory methods to create the `SDK` instance later as needed. What do you think?

I see adjusting cache size and timeout more likely options for clients rather than full Cache and OdpManager customazations. Providing that path with factory will be also a good solution too.

jaeopt: > I am thinking of handling disable / enable at `optimizely` level. When enabled, ODP object will be there and it will always work. When disabled, there will be no ODP instance at all. What do you suggest?

I was concerned about "if odpManager == null" checking in many places we need odpManager. We can try wrap those checking into OdpManager. Other than that, disabling at the top level is also a good option too.

core-api/src/main/java/com/optimizely/ab/odp/ODPManager.java
jaeopt: See TDD which defines accepted types for event data.

zashraf1985: Done!

core-api/src/main/java/com/optimizely/ab/odp/ODPManager.java
zashraf1985: I would advise to defer this discussion till the top level PR. We can then see if it looks good at top level other wise i will add this to ODPManager. what do you say?

core-api/src/main/java/com/optimizely/ab/odp/ODPManager.java
zashraf1985: about top level fetch and event methods, `SegmentManager` and `EventManager` provide some overrides. If i add all those overrides, it looks redundant. I thought it made more sense to provide access to segmentManager and eventManager directly through getters to use those methods.

jaeopt (changes requested):
Looks good. A couple of small changes suggested.

core-api/src/main/java/com/optimizely/ab/odp/ODPEvent.java
```diff
@@ -64,4 +65,20 @@ public Map<String, Object> getData() {
     public void setData(Map<String, Object> data) {
         this.data = data;
     }
+
+    @Transient
+    public Boolean isDataValid() {
+        for (Object entry: this.data.values()) {
```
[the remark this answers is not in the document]
zashraf1985: Done

core-api/src/main/java/com/optimizely/ab/odp/ODPEventManager.java
```diff
@@ -85,6 +87,10 @@ public void identifyUser(@Nullable String vuid, String userId) {
     }
 
     public void sendEvent(ODPEvent event) {
+        if (!event.isDataValid()) {
+            logger.error("Error Sending Event: ODP data is not valid");
```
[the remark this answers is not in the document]
zashraf1985: Done

jaeopt (approved):
LGTM

PR #496, opened 2022-12-21
fix: ODP - getQualifiedSegments should return null when fetch fails

zashraf1985:
## Summary
`getQualifiedSegments` always returns empty list when fetch fails. Modified it to return null in case of failure. It will make it easy to differentiate between a fetch failure and a legit empty list response from the backed.

## Test plan
All existing tests pass

## Issues
[FSSDK-8728](https://jira.sso.episerver.net/browse/FSSDK-8728)


jaeopt (approved):
LGTM

jaeopt:
@mnoman09 I see you changed to `createUserContextCopy()` to avoid unnecessary identify event sending for legacy apis.
I found the same bug in swift-sdk as well and also expect the same in other sdks too. Can you add a FSC test case to cover this for all sdks? @msohailhussain

msohailhussain:
> 

We found this case using FSC. :)

PR #497, opened 2022-12-28
fix: Made ODP Event Dispatcher thread a daemon

zashraf1985:
## Summary
The ODP Event Dispatcher thread was not terminating with the application unless closed explicitly. Converted the thread to a daemon so that it stops when the main app exits.

## Test plan
- Manually tested thoroughly
- All existing tests pass

## Issues
[FSSDK-8727](https://jira.sso.episerver.net/browse/FSSDK-8727)


jaeopt (approved):
LGTM

PR #498, opened 2023-01-02
Feat: Added odp config disable functionality

mnoman09:
## Summary
- Added odp config disable functionality
- Fixed getDefaultRequestConfigWithTimeout function to allow creating separate requestConfigs with timeout

## Test plan
All test should pass.

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/odp/ODPManager.java
+++ b/core-api/src/main/java/com/optimizely/ab/odp/ODPManager.java
@@ -36,9 +36,18 @@ public class ODPManager implements AutoCloseable {
     }
 
     private ODPManager(@Nonnull ODPSegmentManager segmentManager, @Nonnull ODPEventManager eventManager) {
-        this.segmentManager = segmentManager;
-        this.eventManager = eventManager;
-        this.eventManager.start();
+        this(segmentManager, eventManager, false);
+    }
+
+    private ODPManager(@Nonnull ODPSegmentManager segmentManager, @Nonnull ODPEventManager eventManager, boolean disable) {
+        if (!disable) {
+            this.segmentManager = segmentManager;
+            this.eventManager = eventManager;
+            this.eventManager.start();
+        } else {
+            this.segmentManager = null;
+            this.eventManager = null;
+        }
     }
 
     public ODPSegmentManager getSegmentManager() {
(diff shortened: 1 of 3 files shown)


core-api/src/main/java/com/optimizely/ab/odp/ODPManager.java
```diff
@@ -36,9 +36,18 @@ private ODPManager(@Nonnull ODPApiManager apiManager) {
...
 
     private ODPManager(@Nonnull ODPSegmentManager segmentManager, @Nonnull ODPEventManager eventManager) {
-        this.segmentManager = segmentManager;
-        this.eventManager = eventManager;
-        this.eventManager.start();
+        this(segmentManager, eventManager, false);
+    }
+
+    private ODPManager(@Nonnull ODPSegmentManager segmentManager, @Nonnull ODPEventManager eventManager, boolean disable) {
+        if (!disable) {
```
jaeopt, line 43: Java integrates ODP with builder pattern. ODPManager is expected to pass for initialization only when enabled.
I think we may not need this redundant "disable" control here.

enabled: 
```
    OptimizelyBuilder().withOdpManager(...).build()
```
disabled:
```
    OptimizelyBuilder().build()
```

core-httpclient-impl/src/main/java/com/optimizely/ab/HttpClientUtils.java
```diff
@@ -39,13 +39,11 @@ private HttpClientUtils() {
         .build();
 
     public static RequestConfig getDefaultRequestConfigWithTimeout(int timeoutMillis) {
-        if (requestConfigWithTimeout == null) {
```
jaeopt, line 42: I understand we try to re-use RequestConfig once created. Curious why we do not want to reuse?

jaeopt (changes requested):
A few clarifications. We can continue discussion in Teams if I miss the points.

core-httpclient-impl/src/main/java/com/optimizely/ab/HttpClientUtils.java
mnoman09: this function is getting used at multiple places (for setting request config with timeout of SegmentAPIManager as well as for EventAPIManager). If we want to use different timeout for different Request config then this check will not allow that as it will return the first object that was initialized.

PR #500, opened 2023-01-10
Removed setting up batchSize and it will set it to 1 when you set flushInterval to 0

mnoman09:
## Summary
- Removes support for changing odp batchSize.
- If flush interval is set to 0, flushEvent timer is not started and batchSize is set to 1 to dispatch events immediately .

## Test plan
- Fixed unit tests 
- All previous tests should pass.


jaeopt (approved):
LGTM

jaeopt:
Forced merged since conflict with testapp fix.

PR #501, opened 2023-01-11
Fix: Added notificationRegistry to make sure that odpSettings updates

mnoman09:
## Summary
- NotificationRegistry is added to create default NotificationCenter for internal use for updating ODPConfigSetting. 
- This is to handle a corner case that if in any case user passes different notificationCenter to Optimizely and ProjectconfigManager then the Notification of UpdateProjectConfig will not trigger because its handler is getting added in the main NotificationCenter object. 
- This will make sure UpdateProjectConfig will update ODPConfigSetting everytime irrespective of different NotificationCenter passed by user.
- Added getCachedConfig method in ProjectconfigManager to return ProjectConfig without await. 

**NOTE: breaking change**
Anyone who is implementing custom ProjectConfigManager should now have to implement `getCachedConfig` and `getSdkKey `. 

## Test plan
- Added additional unit test to test NotificationRegistry 
- All tests should pass.

## Issue
[FSSDK-8769](https://jira.sso.episerver.net/browse/FSSDK-8769)

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/Optimizely.java
+++ b/core-api/src/main/java/com/optimizely/ab/Optimizely.java
@@ -129,7 +129,7 @@ public class Optimizely implements AutoCloseable {
                 updateODPSettings();
             }
             if (projectConfigManager != null) {
-                NotificationRegistry.getNotificationCenter(projectConfigManager.getSDKKey()).
+                NotificationRegistry.getInternalNotificationCenter(projectConfigManager.getSDKKey()).
                     addNotificationHandler(UpdateConfigNotification.class,
                         configNotification -> { updateODPSettings(); });
             }
--- a/core-api/src/main/java/com/optimizely/ab/Optimizely.java
+++ b/core-api/src/main/java/com/optimizely/ab/Optimizely.java
@@ -125,10 +125,10 @@ public class Optimizely implements AutoCloseable {
 
         if (odpManager != null) {
             odpManager.getEventManager().start();
-            if (!(projectConfigManager instanceof PollingProjectConfigManager)) {
+            if (projectConfigManager.getCachedConfig() != null) {
                 updateODPSettings();
             }
-            if (projectConfigManager != null && projectConfigManager.getSDKKey() != null) {
+            if (projectConfigManager.getSDKKey() != null) {
                 NotificationRegistry.getInternalNotificationCenter(projectConfigManager.getSDKKey()).
                     addNotificationHandler(UpdateConfigNotification.class,
                         configNotification -> { updateODPSettings(); });
@@ -1485,8 +1485,8 @@ public class Optimizely implements AutoCloseable {
     }
 
     private void updateODPSettings() {
-        if (odpManager != null && getProjectConfig() != null) {
-            ProjectConfig projectConfig = getProjectConfig();
+        ProjectConfig projectConfig = projectConfigManager.getCachedConfig();
+        if (odpManager != null && projectConfig != null) {
             odpManager.updateSettings(projectConfig.getHostForODP(), projectConfig.getPublicKeyForODP(), projectConfig.getAllSegments());
         }
     }
--- a/core-api/src/main/java/com/optimizely/ab/Optimizely.java
+++ b/core-api/src/main/java/com/optimizely/ab/Optimizely.java
@@ -125,10 +125,10 @@ public class Optimizely implements AutoCloseable {
 
         if (odpManager != null) {
             odpManager.getEventManager().start();
-            if (getProjectConfig() != null) {
+            if (!(projectConfigManager instanceof PollingProjectConfigManager)) {
                 updateODPSettings();
             }
-            if (projectConfigManager != null) {
+            if (projectConfigManager != null && projectConfigManager.getSDKKey() != null) {
                 NotificationRegistry.getInternalNotificationCenter(projectConfigManager.getSDKKey()).
                     addNotificationHandler(UpdateConfigNotification.class,
                         configNotification -> { updateODPSettings(); });
@@ -159,7 +159,8 @@ public class Optimizely implements AutoCloseable {
         tryClose(eventProcessor);
         tryClose(eventHandler);
         tryClose(projectConfigManager);
-        NotificationRegistry.clearNotificationCenterRegistry();
+        notificationCenter.clearAllNotificationListeners();
+        NotificationRegistry.clearNotificationCenterRegistry(projectConfigManager.getSDKKey());
         if (odpManager != null) {
             tryClose(odpManager);
         }
--- a/core-api/src/main/java/com/optimizely/ab/Optimizely.java
+++ b/core-api/src/main/java/com/optimizely/ab/Optimizely.java
@@ -27,6 +27,7 @@ import com.optimizely.ab.error.NoOpErrorHandler;
 import com.optimizely.ab.event.*;
 import com.optimizely.ab.event.internal.*;
 import com.optimizely.ab.event.internal.payload.EventBatch;
+import com.optimizely.ab.internal.NotificationRegistry;
 import com.optimizely.ab.notification.*;
 import com.optimizely.ab.odp.*;
 import com.optimizely.ab.optimizelyconfig.OptimizelyConfig;
@@ -127,7 +128,12 @@ public class Optimizely implements AutoCloseable {
             if (getProjectConfig() != null) {
                 updateODPSettings();
             }
-            addUpdateConfigNotificationHandler(configNotification -> { updateODPSettings(); });
+            if (projectConfigManager != null) {
+                NotificationRegistry.getNotificationCenter(projectConfigManager.getSDKKey()).
+                    addNotificationHandler(UpdateConfigNotification.class,
+                        configNotification -> { updateODPSettings(); });
+            }
+
         }
     }
 
@@ -153,6 +159,7 @@ public class Optimizely implements AutoCloseable {
         tryClose(eventProcessor);
         tryClose(eventHandler);
         tryClose(projectConfigManager);
+        NotificationRegistry.clearNotificationCenterRegistry();
         if (odpManager != null) {
             tryClose(odpManager);
         }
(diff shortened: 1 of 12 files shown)


core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -127,7 +128,12 @@ private Optimizely(@Nonnull EventHandler eventHandler,
             if (getProjectConfig() != null) {
                 updateODPSettings();
             }
-            addUpdateConfigNotificationHandler(configNotification -> { updateODPSettings(); });
+            if (projectConfigManager != null) {
+                NotificationRegistry.getNotificationCenter(projectConfigManager.getSDKKey()).
```
jaeopt, line 132: This can be confusing with with public NotificationCenter. What about changing names to 
`NotificationRegistry.getInternalNotificationCenter` or some other names to avoid confusion?
@msohailhussain What do you think?

core-api/src/main/java/com/optimizely/ab/config/ProjectConfigManager.java
```diff
@@ -23,5 +23,7 @@ public interface ProjectConfigManager {
      * @return ProjectConfig
      */
     ProjectConfig getConfig();
+
+    String getSDKKey();
```
jaeopt, line 27: Can we get sdkKey via `getConfig().getSDKKey()` instead of adding the new inferface?
This will be convenient but we want to avoid any breaking changes with this new method here.

core-httpclient-impl/src/main/java/com/optimizely/ab/config/HttpProjectConfigManager.java
```diff
@@ -73,11 +74,13 @@ private HttpProjectConfigManager(long period,
                                      String datafileAccessToken,
                                      long blockingTimeoutPeriod,
                                      TimeUnit blockingTimeoutUnit,
-                                     NotificationCenter notificationCenter) {
+                                     NotificationCenter notificationCenter,
+                                     String sdkKey) {
```
jaeopt, line 78: Can we also avoid this change as well if we can get sdkKey via ProjectConfig?

jaeopt (changes requested):
Looks good. Some changes suggested.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
mnoman09, line 132: @jaeopt we are getting sdkKey from projectConfigManager instead of getProjectConfig() because of Asynchronous call of httpConfigManager it might not able to add this NotificationHandler when projectConfig is not fetched.

core-api/src/main/java/com/optimizely/ab/config/ProjectConfigManager.java
mnoman09: see [this](https://github.com/optimizely/java-sdk/pull/501/files#r1067378831) comment. It will make sure that even when the config is not fetched, notificationCenter will contain the handler, which updates the ODPConfig.

jaeopt: @mnoman09 Got it! Can we add a default implementation of this method (returning null), so no changes required for existing implementations. They'll all share the same internal notificationCenter for "null" sdkKey (until they fix it), which looks safe to me.
I believe the default interface method is supported in Java8 and also supported in android.

core-httpclient-impl/src/main/java/com/optimizely/ab/config/HttpProjectConfigManager.java
mnoman09: Same as above we need to have sdkKey pass in here, to make sure that the when we call getSDKKey on projectConfigManager then we won't get it null in case if it is http config Manager.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -153,6 +159,7 @@ public void close() {
         tryClose(eventProcessor);
         tryClose(eventHandler);
         tryClose(projectConfigManager);
+        NotificationRegistry.clearNotificationCenterRegistry();
```
jaeopt, line 162: Is this safe to clear all notificationCenters? We can clear one for the current sdkKey. @msohailhussain also shared concern about memory leak when we close a notificationCenter, which may be addressed together when we clean it up.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -127,7 +128,12 @@ private Optimizely(@Nonnull EventHandler eventHandler,
             if (getProjectConfig() != null) {
```
jaeopt, line 128: It looks like my previous comment on this line got lost :)
We discussed this blocking "getProjectConfig" while constructing Optimizely is not desirable, so we need to make it return null if not available. @msohailhussain may have a good idea on how to fix it.

jaeopt (changes requested):
All changes look great. A couple of more changes suggested.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -124,10 +125,15 @@ private Optimizely(@Nonnull EventHandler eventHandler,
 
         if (odpManager != null) {
             odpManager.getEventManager().start();
-            if (getProjectConfig() != null) {
+            if (!(projectConfigManager instanceof PollingProjectConfigManager)) {
```
jaeopt, line 128: PollingProjectConfigManager can also have a preset projectConfig from a fallback datafile. In that case, we have to call `updateODPSettings` for the projectConfig immediately (we may not get notification later if the fallback is updated already).
I think we can keep the `getProjectConfig() != null` condition as before, with the change that `getProjectConfig() returns null immediately if config is not available.

jaeopt (changes requested):
We may need one more change. See my comments.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
mnoman09: `getProjectConfig() != null`  condition is not possible without wait, with our current implementation. However, we can use the similar approach as we are using in Csharp, adding additional function of `getCachedConfig()` in `ProjectConfigManager`. Now the only concern I have, related to this approach, is that do we want to enforce its implementation or return null by default as we are doing for `getSDKKey()` function? 
In my opinion we should enforce it because if anyone uses custom `ProjectConfigManager` without giving `getCachedConfig` implementation then they won't be able use ODP as it only gets config using this new function.

jaeopt: @mnoman09 Agreed. Let's make them (`getSdkKey` and `getCachedConfig`) mandatory changes and keep them in the list of breaking changes for release.

core-httpclient-impl/src/main/java/com/optimizely/ab/config/HttpProjectConfigManager.java
```diff
@@ -357,7 +365,8 @@ public HttpProjectConfigManager build(boolean defer) {
                 datafileAccessToken,
                 blockingTimeoutPeriod,
                 blockingTimeoutUnit,
-                notificationCenter);
+                notificationCenter,
+                sdkKey);
```
jaeopt, line 369: For existing clients using ".withUrl()" method, `sdkKey` can be null here.
What about moving lines 345-347 to 343, so we clarify a breaking change to those clients, instead of letting them debug why ODP not working?

jaeopt (changes requested):
All changes look good to me!
A couple of small changes suggested - clarifications about required changes for existing custom configMangers. Android-sdk is one of them with tricky changes required.

jaeopt:
@mnoman09 It'll be great if we can have a note about these breaking changes so we do not miss them when releasing a major rev.
We can use the "Unreleased" header in `CHANGELOG.md` like csharp-sdk - https://github.com/optimizely/csharp-sdk/blob/master/CHANGELOG.md

core-httpclient-impl/src/main/java/com/optimizely/ab/config/HttpProjectConfigManager.java
```diff
@@ -367,7 +366,7 @@ public HttpProjectConfigManager build(boolean defer) {
                 blockingTimeoutUnit,
                 notificationCenter,
                 sdkKey);
-
+            httpProjectManager.setSdkKey(sdkKey);
```
jaeopt, line 369: Isn't this `seSdkKey` is redundant? I see `sdkKey' parameter to the constructor (line 368).

core-api/src/main/java/com/optimizely/ab/config/PollingProjectConfigManager.java
```diff
@@ -203,6 +208,10 @@ public synchronized void close() {
         started = false;
     }
 
+    protected void setSdkKey(String sdkKey) {
+        this.sdkKey = sdkKey;
+    }
+
```
jaeopt, line 214: What about moving `getSdkKey` from HttpProjectConfigManager to here? It make look better with get/set combo here and HttpProjectConfigManager inherit both.

jaeopt (approved):
LGTM - See a couple of small changes to clean up.

core-httpclient-impl/src/main/java/com/optimizely/ab/config/HttpProjectConfigManager.java
mnoman09: This will ensure that to trigger notification user provided sdkKey will always be prioritized, instead of what is available in ProjectConfig. This allows registry to always trigger notification even if the SDKKey is changed in the Projectconfig.

jaeopt: @mnoman09 If we need this extra description for proper initializing sdkKey, this may be a source of issues. If you have to call setSdkKey() for proper overriding, can we remove sdkKey from constructor?

mnoman09: I already removed it from the HttpProjectConfigManager constructor if that is what you are asking?

jaeopt: @mnoman09 looks good!

PR #503, opened 2023-02-02
Refact: ODPAPIManager now returns fetchQualifiedSegments list instead of string

mnoman09:
## Summary
- Refactored code of ODPAPIManager fetchQualifiedSegments  function to return list of qualifiedSegments. This is to keep it consistent with other SDKs
- This allows user to override the function and return the qualifiedSegment and they won't have to figure out how to parse the graphql response. 

## Test plan
Fixed some unit tests according to this new change. All unit tests should pass. 

## Issues
- [FSSDK-8842](https://jira.sso.episerver.net/browse/FSSDK-8842)

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/odp/ODPEventManager.java
+++ b/core-api/src/main/java/com/optimizely/ab/odp/ODPEventManager.java
@@ -107,7 +107,7 @@ public class ODPEventManager {
         if (vuid != null) {
             identifiers.put(ODPUserKey.VUID.getKeyString(), vuid);
         }
-        if (userId != null) {
+        if (userId != null && !isVuid(userId)) {
             identifiers.put(ODPUserKey.FS_USER_ID.getKeyString(), userId);
         }
         ODPEvent event = new ODPEvent("fullstack", "identified", identifiers, null);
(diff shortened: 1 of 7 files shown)


jaeopt (approved):
LGTM

core-api/src/main/java/com/optimizely/ab/odp/ODPEventManager.java
```diff
@@ -107,7 +107,7 @@ public void identifyUser(@Nullable String vuid, @Nullable String userId) {
         if (vuid != null) {
             identifiers.put(ODPUserKey.VUID.getKeyString(), vuid);
         }
-        if (userId != null) {
+        if (userId != null && !isVuid(userId)) {
             identifiers.put(ODPUserKey.FS_USER_ID.getKeyString(), userId);
         }
```
jaeopt, line 112: This is a great catch! We also need to consider to add vuid as userId (android-sdk passes it as a normal userId). 
I'll add new test cases into android-sdk.

```suggestion
        if (userId != null) {
            if (isVuid(userId)) {
                        identifiers.put(ODPUserKey.VUID.getKeyString(), userId);
            } else {
                        identifiers.put(ODPUserKey.FS_USER_ID.getKeyString(), userId);
            }
        }
```

jaeopt (changes requested):
A fix suggested to support android-sdk.

jaeopt (approved):
LGTM

PR #504, opened 2023-02-09
fix(ats): convert graphQL query for fetchQualifiedSegments with vuid

jaeopt:
## Summary
When vuid is used as userId, graphQL fetch segments api should be called with "vuid" as a query identifier.

## Test plan
- Unit test validating "vuid" as a query identifier.

## Issues
- [FSSDK-8383](https://jira.sso.episerver.net/browse/FSSDK-8383)


mnoman09 (approved):
LGTM

PR #509, opened 2023-03-13
fix(odp): check odp identifiers not empty before sending

jaeopt:
## Summary
Check if ODP events have at least one key-value pair for identifiers.
If not, discard the request without sending to the ODP server.

## Test plan
- Add a test with empty identifiers.

## Issues
- [FSSDK-8947](https://jira.sso.episerver.net/browse/FSSDK-8947)

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/Optimizely.java
+++ b/core-api/src/main/java/com/optimizely/ab/Optimizely.java
@@ -1468,6 +1468,15 @@ public class Optimizely implements AutoCloseable {
         return odpManager;
     }
 
+
+    /**
+     * Send an event to the ODP server.
+     *
+     * @param type the event type (default = "fullstack").
+     * @param action the event action name.
+     * @param identifiers a dictionary for identifiers. The caller must provide at least one key-value pair.
+     * @param data a dictionary for associated data. The default event data will be added to this data before sending to the ODP server.
+     */
     public void sendODPEvent(@Nullable String type, @Nonnull String action, @Nullable Map<String, String> identifiers, @Nullable Map<String, Object> data) {
         if (odpManager != null) {
             ODPEvent event = new ODPEvent(type, action, identifiers, data);
--- a/core-api/src/main/java/com/optimizely/ab/odp/ODPEvent.java
+++ b/core-api/src/main/java/com/optimizely/ab/odp/ODPEvent.java
@@ -84,4 +84,10 @@ public class ODPEvent {
         }
         return true;
     }
+
+    @Transient
+    public Boolean isIdentifiersValid() {
+        return !identifiers.isEmpty();
+    }
+
 }
--- a/core-api/src/main/java/com/optimizely/ab/odp/ODPEventManager.java
+++ b/core-api/src/main/java/com/optimizely/ab/odp/ODPEventManager.java
@@ -119,12 +119,19 @@ public class ODPEventManager {
     }
 
     public void sendEvent(ODPEvent event) {
-        if (!event.isDataValid()) {
-            logger.error("ODP event send failed (ODP data is not valid)");
-            return;
-        }
         event.setData(augmentCommonData(event.getData()));
         event.setIdentifiers(augmentCommonIdentifiers(event.getIdentifiers()));
+
+        if (!event.isIdentifiersValid()) {
+            logger.error("ODP event send failed (event identifiers must have at least one key-value pair)");
+            return;
+        }
+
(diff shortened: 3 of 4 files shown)


core-api/src/main/java/com/optimizely/ab/Optimizely.java
```diff
@@ -1468,6 +1468,15 @@ public ODPManager getODPManager() {
...
+
+    /**
+     * Send an event to the ODP server.
+     *
+     * @param type the event type (default = "fullstack").
+     * @param action the event action name.
+     * @param identifiers a dictionary for identifiers. The caller must provide at least one key-value pair.
+     * @param data a dictionary for associated data. The default event data will be added to this data before sending to the ODP server.
+     */
     public void sendODPEvent(@Nullable String type, @Nonnull String action, @Nullable Map<String, String> identifiers, @Nullable Map<String, Object> data) {
```
mnoman09, line 1480: It think we should change annotation of identifiers to Nonnull, if we are expecting that caller must provide at least one key value pair.

core-api/src/main/java/com/optimizely/ab/odp/ODPEventManager.java
```diff
@@ -119,12 +119,19 @@ public void identifyUser(@Nullable String vuid, @Nullable String userId) {
     }
 
     public void sendEvent(ODPEvent event) {
+        event.setData(augmentCommonData(event.getData()));
```
mnoman09, line 122: is there a reason why we move this before isDataValid check?

mnoman09 (changes requested):
Overall looks good added few comments.

core-api/src/main/java/com/optimizely/ab/odp/ODPEventManager.java
jaeopt: So, we can add extra protection against "commonData" injected from clients (android-sdk, etc) as well.

core-api/src/main/java/com/optimizely/ab/odp/ODPEvent.java
```diff
@@ -84,4 +84,10 @@ public Boolean isDataValid() {
         }
         return true;
     }
+
+    @Transient
+    public Boolean isIdentifiersValid() {
+        return !identifiers.isEmpty();
```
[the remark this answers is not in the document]
jaeopt: This `identifiers` is instantiated in the constructor. I add Nonnull annotations to make clear.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
jaeopt: @mnoman09 It turns out that this nullable option is needed to support Android (which sets non-empty common identifiers when configuring the java core). I added this clarification into the doc string. 
Another option was to add "vuid" identifier in android-sdk before passing down to java-sdk, but it was not clean since some implicit events ("identifyUser") are created in the java core.

mnoman09: But we are going to make this change in android as well right? Or android had to pass identifiers as null?

core-api/src/main/java/com/optimizely/ab/odp/ODPEvent.java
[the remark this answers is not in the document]
mnoman09: A null check is still necessary because annotation will just give warning in some cases and null pointer exception can cause issue.

core-api/src/main/java/com/optimizely/ab/Optimizely.java
jaeopt: Android (and other client-sdks) customers are allowed to call with empty since SDK is supposed to add at least one identifiers (vuid).

core-api/src/main/java/com/optimizely/ab/odp/ODPEvent.java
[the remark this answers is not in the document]
jaeopt: @mnoman09 I may be confused, but do we still need null checking even with 
```
    @Nonnull private Map<String, String> identifiers;

     this.identifiers = identifiers != null ? identifiers : Collections.emptyMap();
```

core-api/src/main/java/com/optimizely/ab/Optimizely.java
mnoman09: Sounds good.

mnoman09 (approved):
Lgtm

PR #511, opened 2023-04-19
[FSSDK-9054] fix(odp): flush odp events instantly upon calling close

mnoman09:
## Summary
- Added shutdown signal instead of Boolean variable, because previously event queue was waiting for flushInterval to meet before shutting down. 
- Fixed unit test which was checking for flushing events upon calling stop. 

## Test plan
All tests should pass.

## Issues
- [FSSDK-9054](https://jira.sso.episerver.net/browse/FSSDK-9054)


jaeopt (approved):
LGTM

PR #514, opened 2023-05-02
[FSSDK-9127] fix: Handle ODP INVALID_IDENTIFIER_EXCEPTION code

mnoman09:
## Summary

Support a new format for ODP `InvalidIdentifierException` (when SDK fetches segments with a non-registered userId).

```
       "extensions": {
            "code": "INVALID_IDENTIFIER_EXCEPTION",
            "classification": "DataFetchingException"
       }
```

## Test plan
- updated unit tests to cover the new response format.
- Existing unit tests should pass
- FSC should continue to pass

## Issues
- FSSDK-9127

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/odp/parser/impl/GsonParser.java
+++ b/core-api/src/main/java/com/optimizely/ab/odp/parser/impl/GsonParser.java
@@ -1,5 +1,5 @@
 /**
- *    Copyright 2022, Optimizely Inc. and contributors
+ *    Copyright 2022-2023, Optimizely Inc. and contributors
  *
  *    Licensed under the Apache License, Version 2.0 (the "License");
  *    you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


core-api/src/main/java/com/optimizely/ab/odp/parser/impl/GsonParser.java
```diff
@@ -34,15 +34,16 @@ public List<String> parseQualifiedSegments(String responseJson) {
...
-                    if (i > 0) {
-                        logMessage.append(", ");
+                JsonArray errors = root.getAsJsonArray("errors");
+                JsonObject extensions = errors.get(0).getAsJsonObject().get("extensions").getAsJsonObject();
+                if (extensions != null && extensions.has("code")) {
+                    if (extensions.get("code").getAsString().equals("INVALID_IDENTIFIER_EXCEPTION")) {
+                        logger.warn("Audience segments fetch failed (invalid identifier)");
+                    } else {
+                        String errorMessage = extensions.get("classification") == null ? "decode error" : extensions.get("classification").getAsString();
+                        logger.error("Audience segments fetch failed (" + errorMessage + ")");
```
jaeopt, line 44: it's a common case that extensions has "classification" but no "code".
We should log error message for that too and re-use this code for that case.

core-api/src/test/java/com/optimizely/ab/odp/parser/ResponseJsonParserTest.java
```diff
@@ -84,9 +84,33 @@ public void returnNullWhenJsonIsMalformed() {
...
+        assertEquals(null, parsedSegments);
+    }
+
+    @Test
+    public void returnNullAndLogCorrectErrorWhenErrorResponseIsReturnedButCodeValueIsNotInvalidIdentifierException() {
+        String responseToParse = "{\"errors\":[{\"message\":\"Exception while fetching data (/customer) : java.lang.RuntimeException: could not resolve _fs_user_id = wrong_id\",\"locations\":[{\"line\":2,\"column\":3}],\"path\":[\"customer\"],\"extensions\":{\"code\":\"OTHER_EXCEPTIONS\", \"classification\":\"DataFetchingException\"}}],\"data\":{\"customer\":null}}";
+        List<String> parsedSegments =  jsonParser.parseQualifiedSegments(responseToParse);
+        logbackVerifier.expectMessage(Level.ERROR, "Audience segments fetch failed (DataFetchingException)");
+        assertEquals(null, parsedSegments);
+    }
```
jaeopt, line 106: Can we add a test case for extensions without "code" ("classification" only)?

jaeopt (changes requested):
Wow, it's quite expensive to support it in java-sdk. I see a couple of more cases to be covered.

core-api/src/test/java/com/optimizely/ab/odp/parser/ResponseJsonParserTest.java
mnoman09: I have already added this Test case just added error log verification in that test case.

jaeopt (approved):
LGTM

PR #518, opened 2023-06-05
[FSSDK-9422] fix(AsyncEventHandler): add evict timeout to logx connections

jaeopt:
## Summary
Sporadic logx event discards reported. A staled connections after a long idle time may be one of the culprits.
Max connection idle time of 1min is added to AsyncEventHandler, after which connection is forced-closed.

## Test plan
- pass all existing tests.

## Issues
- FSSDK-9422


mnoman09 (approved):
LGTM

PR #524, opened 2023-06-16
[FSSDK-9432] fix: fix to support arbitrary client names to be included in logx and odp events.

jaeopt:
## Summary

1. Fix to support arbitrary client names to be included in logx and odp events. 

- The previous implementation allows only a fixed set of values (java-sdk, android-sdk, and android-tv-sdk) with enum. 
- We need to make it more flexible to support other sdks too (flutter, react-native, etc) wrapping the android-sdk and java-sdk cores.
- Old methods for enum access are all deprecated.

2. Downgrade jackson library version down to 2.13.5 (2.14+ uses Java8 langs not supported in Android)

## Test plan
Add unit tests for new APIs.
Pass all existing tests.

## Issues
- FSSDK-9432


mnoman09 (approved):
LGTM

PR #540, opened 2024-03-19
Align to use virtual threads

### LINKED ISSUE #537: [ENHANCEMENT] [question] Is optimizely virtual thread / loom friendly?
### Description

I wonder if optimizely is loom-friendly - e.g. does not pin the platform threads? I can see that you hardcode underneath platform thread, I would like to replace it with virtual one, currently I can do it only with nasty reflection

### Benefits

Users of sdk can use its resources effectively. The very same goes for reactive apps.

### Detail

_No response_

### Examples

_No response_

### Risks/Downsides

_No response_

wyhasany:
* In scope of this ticket I removed synchronized blocks as they pin virtual thread to carrier thread.
* Expose user way to inject ThreadFactory for creating virtual threads

## Issues
- Fixes #537

### DIFF
--- a/core-httpclient-impl/src/main/java/com/optimizely/ab/event/AsyncEventHandler.java
+++ b/core-httpclient-impl/src/main/java/com/optimizely/ab/event/AsyncEventHandler.java
@@ -21,6 +21,8 @@ import com.optimizely.ab.OptimizelyHttpClient;
 import com.optimizely.ab.annotations.VisibleForTesting;
 
 import com.optimizely.ab.internal.PropertyUtils;
+import java.util.concurrent.ThreadFactory;
+import javax.annotation.Nullable;
 import org.apache.http.HttpResponse;
 import org.apache.http.client.ClientProtocolException;
 import org.apache.http.client.ResponseHandler;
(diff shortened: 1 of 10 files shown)


wyhasany:
@jaeopt just done it

core-httpclient-impl/src/main/java/com/optimizely/ab/event/AsyncEventHandler.java
```diff
@@ -123,10 +136,17 @@ public AsyncEventHandler(int queueCapacity,
             .withEvictIdleConnections(1L, TimeUnit.MINUTES)
             .build();
 
+        NamedThreadFactory namedThreadFactory;
+        if (threadFactory == null) {
+            namedThreadFactory = new NamedThreadFactory("optimizely-event-dispatcher-thread-%s", true);
+        } else {
+            namedThreadFactory = new NamedThreadFactory("optimizely-event-dispatcher-thread-%s", true, threadFactory);
+        }
```
jaeopt, line 144:
```suggestion
        NamedThreadFactory namedThreadFactory = new NamedThreadFactory("optimizely-event-dispatcher-thread-%s", true, threadFactory);
```

jaeopt (changes requested):
Looks good! A couple of clean up suggested to be consistent with other parts.

core-httpclient-impl/src/main/java/com/optimizely/ab/event/AsyncEventHandler.java
jaeopt: @wyhasany a question - replacing all "synchronized" with "lock" is required for Loom virtual threads or just for performance improvement?

wyhasany: it is required for loom, as loom can't switch virtual thread when it is blocked by synchronized block.
> There are two scenarios in which a virtual thread cannot be unmounted during blocking operations because it is >pinned to its carrier:
>
>   1. When it executes code inside a synchronized block or method, or...
https://openjdk.org/jeps/444

wyhasany:
@jaeopt Thank you for review. I applied all yours suggestions

jaeopt:
@wyhasany Thanks for your contribution. It's merged and will be included in the next release (next week).

PR #602, opened 2026-04-03
[AI-FSSDK] [FSSDK-12418] Remove experiment type validation from config parsing

FarhanAnjum-opti:
## Summary

Removes experiment type validation from config parsing to ensure forward compatibility. Previously, all four config parsers validated the experiment "type" field against a known set of values and rejected the entire datafile if an unknown type was encountered, which would break existing SDK versions when new experiment types are added in the future.

## Changes

- Removed type validation blocks from all four config parsers (JsonConfigParser, JsonSimpleConfigParser, DatafileGsonDeserializer, DatafileJacksonDeserializer)
- Unknown experiment types are now silently accepted and stored as-is
- Added test verifying unknown experiment types are accepted during config parsing

## Jira Ticket
[FSSDK-12418](https://optimizely-ext.atlassian.net/browse/FSSDK-12418)

[FSSDK-12418]: https://optimizely-ext.atlassian.net/browse/FSSDK-12418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core-api/src/test/java/com/optimizely/ab/config/FeatureRolloutConfigTest.java
+++ b/core-api/src/test/java/com/optimizely/ab/config/FeatureRolloutConfigTest.java
@@ -135,7 +135,22 @@ public class FeatureRolloutConfigTest {
     }
 
     /**
-     * Test 6: Type field parsed - experiments with type field in the datafile
+     * Test 6: Unknown type accepted - experiment with type "new_unknown_type"
+     * does NOT cause error or rejection, and config parsing succeeds.
+     */
+    @Test
+    public void unknownExperimentTypeAccepted() {
+        Experiment experiment = projectConfig.getExperimentKeyMapping().get("unknown_type_experiment");
+        assertNotNull("Experiment with unknown type should be parsed successfully", experiment);
+        assertEquals("new_unknown_type", experiment.getType());
+        assertEquals("exp_unknown_type", experiment.getId());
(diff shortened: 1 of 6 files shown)


core-api/src/test/java/com/optimizely/ab/config/FeatureRolloutConfigTest.java
```diff
@@ -135,7 +135,22 @@ public void featureRolloutWithEmptyRolloutIdDoesNotCrash() {
     }
 
     /**
-     * Test 6: Type field parsed - experiments with type field in the datafile
+     * Test 6: Unknown type accepted - experiment with type "new_unknown_type"
+     * does NOT cause error or rejection, and config parsing succeeds.
+     */
+    @Test
+    public void unknownExperimentTypeAccepted() {
```
jaeopt, line 142: @FarhanAnjum-opti  this test validates datafile parsing won't fail because of invalid rule type.
Ideally, we need to validate SDK returns a "valid" decision for the rule ignoring the rule type.
Since we'll cover this full decision/event in FSC, I think we can approve this for all SDKs as is.

jaeopt (approved):
LGTM

PR #513, opened 2023-04-20
[FSSDK-9076] fix: send odp event empty values

mnoman09:
## Summary
- Validate sendOdpEvent empty action and type parameters
- Generate new idempotence_id for every OdpEvent

## Test plan
added tests and all tests should pass.

## Ticket
[FSSDK-9076](https://jira.sso.episerver.net/browse/FSSDK-9076)

PR #629, opened 2026-05-26
[AI-FSSDK] [FSSDK-12670] Block ODP identify event for single identifier

jaeopt:
## Summary

Fixes the ODP identify event to only send when multiple valid identifiers exist. Previously, identify events were always sent even with a single identifier, which provided no linking value and generated unnecessary network traffic.

## Changes
- Changed ODPEventManager.identifyUser to accept a Map of identifiers and filter out null/empty values
- Added guard to skip identify event when fewer than 2 valid identifiers remain
- Updated Optimizely.identifyUser to construct identifier map internally while keeping backward-compatible String parameter

## Jira Ticket
[FSSDK-12670](https://optimizely-ext.atlassian.net/browse/FSSDK-12670)

## Notes
- Reference implementation: Python SDK PR #513

[FSSDK-12670]: https://optimizely-ext.atlassian.net/browse/FSSDK-12670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/core-api/src/main/java/com/optimizely/ab/odp/ODPEventManager.java
+++ b/core-api/src/main/java/com/optimizely/ab/odp/ODPEventManager.java
@@ -139,6 +139,11 @@ public class ODPEventManager {
     }
 
     public void identifyUser(@Nonnull Map<String, String> identifiers) {
+        if (identifiers == null) {
+            logger.debug("ODP identify event is not dispatched (fewer than 2 valid identifiers).");
+            return;
+        }
+
         Map<String, String> validIdentifiers = new HashMap<>();
         for (Map.Entry<String, String> entry : identifiers.entrySet()) {
             if (entry.getValue() != null && !entry.getValue().isEmpty()) {
@@ -147,7 +152,7 @@ public class ODPEventManager {
         }
 
         if (validIdentifiers.size() < 2) {
-            logger.debug("ODP identify event is not dispatched (only one identifier provided).");
+            logger.debug("ODP identify event is not dispatched (fewer than 2 valid identifiers).");
             return;
         }
 
--- a/core-api/src/main/java/com/optimizely/ab/odp/ODPEventManager.java
+++ b/core-api/src/main/java/com/optimizely/ab/odp/ODPEventManager.java
@@ -111,23 +111,21 @@ public class ODPEventManager {
         }
     }
 
-    public void identifyUser(String userId) {
-        identifyUser(null, userId);
-    }
-
-    public void identifyUser(@Nullable String vuid, @Nullable String userId) {
-        Map<String, String> identifiers = new HashMap<>();
-        if (vuid != null) {
-            identifiers.put(ODPUserKey.VUID.getKeyString(), vuid);
-        }
-        if (userId != null) {
-            if (ODPManager.isVuid(userId)) {
-                identifiers.put(ODPUserKey.VUID.getKeyString(), userId);
-            } else {
-                identifiers.put(ODPUserKey.FS_USER_ID.getKeyString(), userId);
+    public void identifyUser(@Nonnull Map<String, String> identifiers) {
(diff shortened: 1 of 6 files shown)


core-api/src/main/java/com/optimizely/ab/odp/ODPEventManager.java
```diff
@@ -127,7 +135,23 @@ public void identifyUser(@Nullable String vuid, @Nullable String userId) {
...
+        Map<String, String> validIdentifiers = new HashMap<>();
+        for (Map.Entry<String, String> entry : identifiers.entrySet()) {
+            if (entry.getValue() != null && !entry.getValue().isEmpty()) {
+                validIdentifiers.put(entry.getKey(), entry.getValue());
+            }
+        }
+
+        if (validIdentifiers.size() < 2) {
+            logger.debug("ODP identify event is not dispatched (only one identifier provided).");
+            return;
```
Mat001, line 158: The log message says "only one identifier provided" but it also fires for zero identifiers (empty map). The test identifyUserSkippedWithEmptyMap verifies this. A more accurate message would be:

logger.debug("ODP identify event is not dispatched (fewer than 2 valid identifiers).");

core-api/src/main/java/com/optimizely/ab/odp/ODPEventManager.java
```diff
@@ -127,7 +135,23 @@ public void identifyUser(@Nullable String vuid, @Nullable String userId) {
                 identifiers.put(ODPUserKey.FS_USER_ID.getKeyString(), userId);
             }
         }
-        ODPEvent event = new ODPEvent("fullstack", "identified", identifiers, null);
+        identifyUser(identifiers);
+    }
+
+    public void identifyUser(@Nonnull Map<String, String> identifiers) {
```
Mat001, line 141: This method annotates @Nonnull but doesn't null-check the parameter. If someone passes null, it will NPE at the for loop. Either add a null guard or trust the annotation — but given the deprecated methods delegate to it, a defensive check would be safer.

Mat001 (changes requested):
Looks pretty good. A few comments.

Additional comment:
Several tests use assertEquals(identifiers.size(), 2) (JUnit convention is assertEquals(expected, actual)). The new test identifyUserSendsWithThreeIdentifiers correctly uses assertEquals(3, eventIdentifiers.size()) but the updated identifyUserWithMultipleIdentifiers still has the args reversed: assertEquals(eventIdentifiers.size(), 2). Inconsistent within the same PR.

core-api/src/main/java/com/optimizely/ab/odp/ODPEventManager.java
```diff
@@ -127,7 +135,28 @@ public void identifyUser(@Nullable String vuid, @Nullable String userId) {
...
+        if (identifiers == null) {
+            logger.debug("ODP identify event is not dispatched (fewer than 2 valid identifiers).");
+            return;
+        }
+
+        Map<String, String> validIdentifiers = new HashMap<>();
+        for (Map.Entry<String, String> entry : identifiers.entrySet()) {
+            if (entry.getValue() != null && !entry.getValue().isEmpty()) {
```
Mat001, line 149: Could we add a comment here, the reason for this condition. > 2 is unintuitive, some comment would be good

Mat001 (approved):
One comment.

core-api/src/main/java/com/optimizely/ab/odp/ODPEventManager.java
jaeopt: Added a comment explaining the rationale. Also applied the same comment across all other SDK PRs for consistency.

